PR #2, opened 2025-07-02
Migrate first ui component

mikiher:
This migrates the components/ui/Btn component. 

The porpose of this PR is mainly to establish basic coding conventions and best practises for the migration of components from vue to tsx (more will likely come as additional components are migrated).

Here's a summary of principles I have for this PR (in no particular order):
- Use `className` property and `mergeClasses` function to allow for passing and merging of additional classes.
- Use `memo`, `useMemo` and `useCallback` when possible for performance
- Be mindful of ARIA and screen-reader support while migrating.
  - Use appropriate ARIA attributes
  - Use the `sr-only` class when appropriate
- Write comprehensive component tests (AI is good for generating those)
  - Fix bugs discovered by tests.

Please let me know if you want to add anything. Some questions that came up while doing this specific component:
- We're using this tailwind anti-pattern of dynamically building tailwind classes. like here:
   ```
   if (paddingX !== undefined) {
      list.push(`px-${paddingX}`)
    }
    if (paddingY !== undefined) {
      list.push(`py-${paddingY}`)
    }
   ```
   I have not touched this when migrating. 
- What should/should not be in tests: For example, I tried to be comprehensive about functionality and edge-cases, but there's no testing for existence of specific tailwind classes, and also no specific testing of proper positioning. Let me know if you have any preferences here.

While migrating, I fixed one specific bug causing a rendering issue when showing a Link. To fix it, I added `inline-flex` to the base class list.


mikiher:
Also Migrated IconBtn (plus tests and component catalog)

mikiher:
Based on our discussion, I've reduced prettier printWidth to from 400 to 160.
Also, it seems I didn't have prettier formatting running at all, so I amde it the default formatter in .vscode/settings.json, and also set Format on Save to true.

I went over the existing files and checked-in the formatted versions.

advplyr:
This is great, I love the components catalog. It made me think about the component management project, [Storybook](https://storybook.js.org/docs), that I mentioned before. We can definitely roll our own light-weight version of this but I think we should have something like it.



> * We're using this tailwind anti-pattern of dynamically building tailwind classes

Let's update these as we see it when migrating. We definitely don't want to keep it in there and as we migrate pages we will update the page using the component.



> * What should/should not be in tests: For example, I tried to be comprehensive about functionality and edge-cases, but there's no testing for existence of specific tailwind classes, and also no specific testing of proper positioning. Let me know if you have any preferences here.

I'm not sure about this. I think what you have is good. Once pages start getting built we can put in more specific tests.

It's the same with coding conventions. We can better define these as we run into them while building.

It already feels much cleaning working in this codebase. Thanks!

advplyr:
I was playing around with Storybook and set up a quick demo https://github.com/audiobookshelf/audiobookshelf-client-react/pull/4

Cursor set up the Btn component and the stories. Let me know what you think. I've never used it before today

PR #4, opened 2025-07-12
Add storybook with btn component

advplyr:
This is testing out [Storybook](https://storybook.js.org/docs) and adds the Btn component with some stories.

To use:
1. npm install
2. npm run storybook

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ yarn-error.log*
 next-env.d.ts
 
 express-server
+*storybook.log
+storybook-static
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,0 +1,23 @@
+import type { StorybookConfig } from "@storybook/nextjs-vite";
+
+const config: StorybookConfig = {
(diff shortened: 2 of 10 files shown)


mikiher:
After spending some time with it and generating stories for all the components I alerady migrated, I think I don't like it so much:
1. Learning curve is a bit on the steep side for me.
2. In the Docs view, it displays stories in fixed-sized iframes, and if the components becomes larger (e.g. opening a menu), you have to scroll down to see the bottom. I tried to find out how to change this behavior, and I couldn't.
3. I haven't even started looking at test integration, but I see some of it isn't free of charge.

Perhaps it's a bit premature, but my gut feeling is that I need to spend more time on the tool than I care to.

With the simple catalog page, I have full control, and it also looks like Cursor is better at generating code for the catalog page than it is at generating stories files.

advplyr:
I'm good with that. Would the catalog page ship in production?

mikiher:
Of course not. It's just for dev purposes.

PR #5, opened 2025-07-17
More component migrations

mikiher:
This migrates additional UI/Widget components.

Components migrated:
- widgets/LoadingSpinner
- ui/ContextMenuDropdown
- ui/Dropdown
- ui/FileInput
- ui/LibraryIcon
- ui/MediaIconPicker
- ui/InputDropdown

In the majority of cases migration also included additional accessibility support, most notably keyboard navigation support and proper aria attributes.

Some general rules I used for accesibility:
- Support arrow navigation (including left/right where it made sense), enter, space, escape, home, end, and tab.
- Prefer aria-activedescendant and visual (non-native) focus on menu and menu items.
  - Tabbing moves between top-level components, and doesn't focus on menus or menu items. 
  - Tab closes a menu if it's open
- Use React's useId for unique id generation.

All migrated components were added to the component catalog.

New packages installed:
- react-transition-group (non-dev): for supporting vue-like transitions
- cypress-real-event (dev): for supprting native event in Cypress on Chromium-based browsers
- cypress-plugin-tab (dev): for adding a missing tab support in Cypress

Additional major changes:
- Refactor menu and menu items in Dropdown and InputDropdown into a new single component DropdownMenu
- Add support for global toast

### DIFF
--- a/src/components/ui/ContextMenuDropdown.tsx
+++ b/src/components/ui/ContextMenuDropdown.tsx
@@ -3,8 +3,6 @@
 import { useState, useRef, useEffect, useCallback, useMemo, useId } from 'react'
 import { useClickOutside } from '@/hooks/useClickOutside'
 import LoadingSpinner from '@/components/widgets/LoadingSpinner'
-import { CSSTransition } from 'react-transition-group'
-import '@/assets/transitions.css'
 import { mergeClasses } from '@/lib/merge-classes'
 
 export interface ContextMenuDropdownSubitem {
@@ -68,7 +66,6 @@ export default function ContextMenuDropdown({
     setFocusedIndex(index)
     setFocusedSubIndex(-1)
     setOpenSubMenuIndex(null)
-    console.log('openMenu', index)
   }, [])
 
   const closeMenu = useCallback(() => {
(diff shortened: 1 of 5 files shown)


src/components/ui/ContextMenuDropdown.tsx
```diff
@@ -0,0 +1,532 @@
...
+      ) : (
+        <div className="h-full w-full flex items-center justify-center">
+          <LoadingSpinner />
+        </div>
+      )}
+
+      <CSSTransition
+        in={showMenu}
+        timeout={{
+          enter: 200,
```
advplyr, line 508: I'm thinking we should have the enter transition 0.
We lose the transition but it feels much snappier and I hardly notice that it isn't transitioning. When I put a few of these side by side with different values the one with 0 felt much better.

advplyr: You're right these aren't working in the nuxt app. My eyes were tricking me.
I did some testing with the transitions and I agree we should remove them from dropdowns.

advplyr:
For the input dropdown with new item creation the `onNewItem` function is called multiple times.

1. Type something in and press enter and new item is called
2. Remove focus from the input and new item is called

image

The nuxt app is checking that `input` and `textInput` are different before emitting `newItem`.

This is why the nuxt app has a 50ms timeout on the blur event from the input.

It feels a bit hacky to put a timeout on blur. There may be a better way to build this.

advplyr:
> Can you explain what the sematics of the `editable=false` and `disabled=false` props together mean from your perspective?
> 
> It looks like the Vue implementation also allows you to focus on the input in this case and to select another item from the menu. It just doesn't allow you to type in the input field. The only other thing it doesn't allow you is to set focus programmatically using the `setFocus` method. I doubt we actually need this prop, as it doesn't make a lot of sense. I think we only need `disabled` and should remove `editable`. Do you concur?

Yeah we can remove that. It's not used anywhere in the Vue project.

PR #6, opened 2025-07-20
ContextMenuDropdown: Fix and refactor menu into ContextMenu.tsx

mikiher:
This fixes and refactors ContextMenuDropdown.

Major changes:
1. Refactored the menu into a separate component `ContextMenu`
2. Fixed some small hovering bugs 
3. Replaced the `setTimout` in the parent mouseleave handler with a more robust layout effect.
    When the mouse leaves the parent of the submenu:
      - We only set appropriate state and register intent to potentially close the submenu
      - All other mouse enter/leave events finish and set appropriate state
      - The layout effect considers the new state and decides whether to close the menu or not.
4. Fixed behavior for empty submenus:
    - Added a disabled "No items" button
    - Fixed keyboard navigation to handle empty submenus correctly
5. Added and updated component tests


advplyr:
Working great, thanks!

PR #12, opened 2025-08-09
Migrate TextAreaInput

mikiher:
This migrates both TextAreaInput and TextAreaWithLabel into a single component TextAreaInput (in the current view code, TextAreaInput is never used directly, just as a component in TextAreaWithLabel).

Notes:
- Accessiblity attributes are properly set
- The `transparent` prop was removed, since it's unused in the vue code and doesn't make a lot of sense.
- The original had an imperative blur() method. I have not implemented it since I'm not sure it's required (it is used once in PodcastDetailsEdit.vue, but I'm not sure if we can't do without it). I prefer not to use imperatives, if not strictly needed.
- I don't dig the `readOnly` styles. They seem similar to `disabled`, but not exactly the same. I (tried to) migrate exactly like in the vue code, but it doesn't make a lot of sense to me.

### DIFF
--- a/src/app/(main)/components_catalog/ComponentExamples.tsx
+++ b/src/app/(main)/components_catalog/ComponentExamples.tsx
@@ -11,6 +11,7 @@ import type { DropdownItem } from '@/components/ui/Dropdown'
 import FileInput from '@/components/ui/FileInput'
 import IconBtn from '@/components/ui/IconBtn'
 import InputDropdown from '@/components/ui/InputDropdown'
+import TextareaInput from '@/components/ui/TextareaInput'
 import LibraryIcon from '@/components/ui/LibraryIcon'
 import LoadingIndicator from '@/components/ui/LoadingIndicator'
 import LoadingSpinner from '@/components/widgets/LoadingSpinner'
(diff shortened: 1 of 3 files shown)


advplyr:
> This migrates both TextAreaInput and TextAreaWithLabel into a single component TextAreaInput (in the current view code, TextAreaInput is never used directly, just as a component in TextAreaWithLabel).

We should merge the `TextInput` and `TextInputWithLabel` into a single component also. I think you mentioned that before.
Then we should have a `labelHidden` prop for cases where you don't want to show a label but still want to pass one in for screen readers.

I've been working with Shopify's Polaris design system recently. I appreciate the level of detail that they put in to all aspects of UI.
I got the `labelHidden` from their text field component:
https://polaris-react.shopify.com/components/selection-and-input/text-field#labeling



> * I don't dig the `readOnly` styles. They seem similar to `disabled`, but not exactly the same. I (tried to) migrate exactly like in the vue code, but it doesn't make a lot of sense to me.

I think it looks good. The only thing is we can make the text lighter like `#ccc` instead of `#aaa`.
Referencing Polaris again, this is how they have it.
image

Textinput background/border is the same color for both readonly and disabled.
This similar to what we have.

-----

On a separate note we should start replacing specific colors like `text-gray-400` with `text-disabled`.
And for the readonly we could have `text-readonly` or `text-secondary` instead of `text-[#aaa]`.
That will help with consistency and we can eventually offer color themes.

That can be tackled separately though.

advplyr:
> So as not to invent another color pallete, can we make it text-gray-300 instead of #ccc? it's #d1d5dc in hex terms.

Yeah I just made that change. But like I mentioned at the bottom of my last post these will all be changed to non-specific colors.

PR #13, opened 2025-08-09
Fix disabled and read-only colors + migrate TextInput

mikiher:
As agreed in #12, this adds named disabled and read-only foreground and background colors.

All components were modified to use the new colors, with some minor additional fixes.


src/components/ui/InputDropdown.tsx
```diff
@@ -243,7 +247,7 @@ export default function InputDropdown({
             value={textInput}
             disabled={disabled}
             tabIndex={disabled ? -1 : 0}
-            className={mergeClasses('h-full w-full bg-transparent px-1 outline-none', disabled ? 'cursor-not-allowed' : '')}
+            className={mergeClasses('h-full w-full bg-transparent px-1 outline-none', disabled ? 'cursor-not-allowed text-disabled' : '')}
```
advplyr, line 250:
```suggestion
            className="h-full w-full bg-transparent px-1 outline-none disabled:cursor-not-allowed disabled:text-disabled"
```

I think we should use the disabled modifier on buttons and inputs. Not really that important but it is a bit cleaner doing it the tailwind way.
It is the same with most of the other components as well. RangeInput is already using the disabled modifier.

mikiher: OK, I'll make those changes.

mikiher: Done.

mikiher:
Also migrated TextInput (merged with TextInputWithLabel, as discussed).

Notes:
- TextInput had many props, many of which I considered unnecessary and removed:
   - `paddingY`, `paddingX` can both be set using either `className` or 'customInputClass`
   - `noSpinner`, `textCenter` can be set using `customInputClass`
   - `onKeyup` does not seem to be used anywhere.
   - 'trimWhitespace` I just found baffling - parents can opt to trim whitespaces themselves once they decide the vaue is ready to be saved. It certainly doesn't need to happen inside TextInput
- Added a `label` prop, which also provides the default value for `aria-label`, and is shown above the input like in other components.
- Adde a `ref` prop to get a forward ref to the input element (instead of defining imperatives, which I don't like).
- Replaced scoped classes with tailwind classes
- Made focus, disabled, and read-only styles consistent with other input components.
- Made background consistent with other input components.

PTAL

advplyr:
> * `paddingY`, `paddingX` can both be set using either `className` or `customInputClass`

These are on Btn too and can be removed.

Looks good to me, thanks!

PR #14, opened 2025-08-17
Component fixes and standardization

mikiher:
In addition to some small fixes, this PR mainly standardizes behaviors of buttons and input-like controls.

It's based on 3 new components:

1. InputWrapper: A wrapper for input-like controls that provides a consistent look and feel.
2. ButtonBase: A base component for buttons that provides a consistent look and feel.
3. Label: A label component that provides a consistent look and feel to components with labels.

The following aspects are standardized:

1. Label styles (font size and weight, padding and margins, block behavior), sizing (width), and layout.
2. InputWrapper and ButtonBase standardize over the following:
   - border: including readonly and disabled states
   - rounded styles (rounded-md)
   - shadow styles (shadow-md)
   - focus styles, including focus-within for the wrapper and focus-visible for buttons
   - layout styles (flex, items-stretch)
   - positioning styles (relative)
   - background styles (including readonly and disabled states)
   - height: supporting small, medium, large, and auto sizes
   - padding (px-2)
   - disabled state cursor (cursor-not-allowed)
   - before pseudo-element for buttons (governing hover background color)
   - tab index
   - aria attributes (aria-label, aria-disabled) for buttons

This allows for a very consistent look and feel. An example for this is shown in the component examples (look for the "Side By Side Controls" example)

### DIFF
--- a/src/components/ui/Checkbox.tsx
+++ b/src/components/ui/Checkbox.tsx
@@ -39,7 +39,7 @@ export default function Checkbox({
   const checkboxId = useId()
 
   const checkboxWrapperClassName = useMemo(() => {
-    const classes = [checkboxBgClass, borderColorClass]
+    const classes = [checkboxBgClass, disabled ? 'border-checkbox-bg-disabled' : borderColorClass]
 
     if (size === 'small') {
       classes.push('w-4 h-4')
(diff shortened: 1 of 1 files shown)


src/components/ui/ContextMenuDropdown.tsx
```diff
@@ -329,15 +332,24 @@ export default function ContextMenuDropdown({
     }))
   }))
 
+  const buttonClass = useMemo(() => {
+    return mergeClasses('rounded-full before:rounded-full', size === 'small' ? 'w-9' : size === 'large' ? 'w-11' : 'w-10', className)
```
advplyr, line 336: I think the context menu should have an option to not use the button style at all. And when using the button style we should use the same radius as the default icon btn.

For example we have a context menu on the lazy book card
image

And we have on the library item page
image

That is using the horizontal more icon for no particular reason. It would be fine with `more_vert` also.

PR #15, opened 2025-08-19
Migrate TimePicker (now DurationPicker)

mikiher:
This migrates the TimePicker component, changing its name to DurationPicker.

Notable changes:
- Added accessibility support
- Simplified some of the calculations

### DIFF
--- a/src/app/(main)/components_catalog/ComponentExamples.tsx
+++ b/src/app/(main)/components_catalog/ComponentExamples.tsx
@@ -23,6 +23,7 @@ import MultiSelectDropdown from '@/components/ui/MultiSelectDropdown'
 import RangeInput from '@/components/ui/RangeInput'
 import ReadIconBtn from '@/components/ui/ReadIconBtn'
 import TwoStageMultiSelect, { TwoStageMultiSelectContent } from '@/components/ui/TwoStageMultiSelect'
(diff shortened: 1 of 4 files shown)


advplyr:
I made a mistake when building that component on Vue by not thinking about mobile. When not using a native HTML input the onscreen keyboard does not show.

https://github.com/advplyr/audiobookshelf/issues/1776

We need to find a way to switch those divs to inputs or put focus on hidden inputs.

mikiher:
Are we married to the single digit spin buttons? It is not very
accessibilty-friendly. Can I suggest a 3 input ([h]hh:mm:ss) design that
would work on both desktop and mobile. Each input would still be a spin
button but would go over the full unit (hours/minutes/seconds) values,
instead of digit by digit. I think that the digit by digit inputs don't
provide a lot of value on desktop, and they're too complicated for mobile.

advplyr:
> Are we married to the single digit spin buttons

Not at all. I'm not following exactly what you mean but I think go for it

mikiher:
Let me implement, and then you can play with it and comment.

mikiher:
OK, PTAL again. 

I'll let you have the pleasure of discovering how it operates by yourself. I think it's pretty inuitive, still accessible, and works on mobile.

PR #16, opened 2025-08-21
Toggle buttons and switches

mikiher:
This migrates ToggleBtns (renamed to ToggleButtonGroup), and ToggleSwitch (in addition to a couple of small unrelated fixes)

The examples are pretty self-explanatory.

As usual, keyboard navigation and accessibility added and checked.


advplyr:
Looks great. I started making some updates to the colors.

Updated the border for input, switch & button base to global variable.
Set up a global variable for the toggle button selected color.

I can go through and try to standardize & move all colors to global variables in a separate PR. As we discussed we want to be able to put in a light theme or high contrast theme by only updating the CSS variables. We're still using `text-white`, `gray`, `black`, etc. in a bunch of places.

PR #17, opened 2025-08-24
Tooltip: Migrate from Vue

mikiher:
This migrates Tooltip. 

As always, accessibilty and keyboard navigation are added:
- Tooltip is displayed/removed when children are focused/blurred
- Escape dismisses tooltip
- aria-describedby added to the first element child.
- role="tooltip" added to the tooltip div

The tooltip can be created in a portal, but by default it's not, unless it is inside a modal.

We have lots  of (i) icons in the Vue code. The ARIA guide says:
> The tooltip is not the appropriate role for the more information "i" icon, ⓘ. A tooltip is directly associated with the owning element. The ⓘ isn't 'described by' detailed information; the tool or control is.

I will create a separate component for (i) icons that doesn't have tooltip semantics (or modify this one by adding some prop).

### DIFF
--- a/src/app/(main)/components_catalog/ComponentExamples.tsx
+++ b/src/app/(main)/components_catalog/ComponentExamples.tsx
@@ -2177,7 +2177,7 @@ export function TooltipExamples() {
         <Example title="Right Edge" className="col-span-1 md:col-span-2 lg:col-span-2">
           <div className="space-y-4 flex flex-col items-end">
             <p className="text-sm text-gray-400 mb-2">Resize to make this button appear close to the right edge of the screen</p>
-            <Tooltip text="right tooltip" position="right">
+            <Tooltip text="Very long tooltip text that should be shifted to the left" position="bottom">
               <Btn>Right</Btn>
             </Tooltip>
           </div>
--- a/src/components/ui/Tooltip.tsx
+++ b/src/components/ui/Tooltip.tsx
@@ -55,8 +55,8 @@ const Tooltip = ({
   const middleware = useMemo(() => {
     const mw = [
       offset(offsetPx),
-      flip({ fallbackAxisSideDirection: 'start' }),
       shift({ padding: edgePadding }),
+      flip({ fallbackAxisSideDirection: 'start' }),
       size({
         padding: edgePadding,
         apply: ({ availableWidth, elements }) => {


advplyr:
Two things:

1. We need tooltip content to wrap with a max width prop.

In some cases the tooltip will be long. For example, on the `LazyBookCard` when the title or subtitle is truncated we use a tooltip to show the full text.

image


There are several settings with longer tooltips also. This may not be good design but it is convenient to put these messages in a tooltip.
Email settings:

image

---

2. We should shift tooltips so that they never overflow the screen.

For example, in the audio player the right-most icon button tooltip is shifted left to prevent overflow. In this case it wouldn't make sense to have the tooltip open to the left and cover the other icon buttons.

image

Side note: we may also want to implement a small arrow that points to the element that opened the tooltip. This isn't important or necessary but is something I saw in [these tooltip guidelines](https://www.nngroup.com/articles/tooltip-guidelines/) and we currently use for the audio player progress bar tooltips.

---

I don't know if this is necessary but another feature that was implemented in the Vue tooltip is to keep the tooltip open when hovering over it. This is the rationale for it: https://github.com/advplyr/audiobookshelf/pull/1188

The response to that PR mentions the case where the tooltip covers another element and so it is a hindrance to keep the tooltip open when hovering over it. https://github.com/advplyr/audiobookshelf/pull/1188#issuecomment-1862116725

[Shopify Polaris tooltip](https://polaris-react.shopify.com/components/overlays/tooltip?example=tooltip-default) handles this with a `dismissOnMouseOut` which I've had to use for that exact scenario.

advplyr:
I just pushed up this example component since I added it for testing but feel free to remove/update it.

image

mikiher:
> 1. We need tooltip content to wrap with a max width prop.

Done. Default is `undefined`.

> 2. We should shift tooltips so that they never overflow the screen.

Done. Floating-ui is now doing a much better job at figuring out positioning and updating it, with its `offset -> flip -> shift -> size` middleware flow.

> Side note: we may also want to implement a small arrow that points to the element that opened the tooltip. 

Done. Floating-ui also helps here with arrow positioning.

> ... keep the tooltip open when hovering over it.

Done. This in accordance with ARIA guidelines for tooltips.

PTAL again.

advplyr:
Here is a case where flip would be good.

image

On a smaller screen and a tooltip opening right. The Vue tooltip logic shifts and covers up the icon but it would be better to flip in this case.

If that floating library has something like flip tolerance that would work. The logic is that we want to flip only when necessary. When necessary would be if shifting would cover the icon like in this example.

mikiher:
So, how about changing the order of shift and flip? 

The order matters. If shift is able to solve clipping issues, then flip would not need to change the position.

advplyr:
> So, how about changing the order of shift and flip?
> 
> The order matters. If shift is able to solve clipping issues, then flip would not need to change the position.

Ah I didn't try that. That sounds like what I'm trying to say

mikiher:
Note, though, that rhe initial positioning of the tooltip still matters even when you swap the order the middlewares.

If you set `position="right"` then it would always flip when close to the edge, because `shift` cannot fix the clipping (it cannot shift over the reference element). So you have to set either `position="top"` or `position="bottom"` - then `shift` would shift it to the right (because it can), and `flip` would decide to leave the position as-is.

That's what I meant when I said "design flaws" - setting `position="right"` when you know it might be close to the right edge is a kind of a design flaw, and `flip` "deals" with it.

PR #22, opened 2025-09-24
CronExpressionBuilder & CronExpressionPreview

mikiher:
This migrates CronExpressionBuilder.vue, and add a CronExpressionPreview component that shows a localized verbal description of the schedule, and a localized next run time.

**CronExpressionBuilder**
The migrated builder is almost a complete re-write of the old one, with the following main differences:
- A fully controlled component
  - Parent sets value and onChange is called on input change
  - All display elements are derived from the value
- No Schedule/Advanced tabs.
  - Instead, "Custom Cron Expression" is just another interval option (appearing last)
- Using the native`type="time"` input for specifying time instead of `Hours` and `Minutes`
- No server API call for validation
  - Validation now happens on every change in the input
  - If value is not valid, border goes red, and error is displayed below
- If client and server time zones disagree, a message is displyed to notify the user that time is specified in the server's time zone.
- No external package dependency for the builder
- Weekday names in the weekdays multi-select are localized using standard Javascript (`Intl`)

**CronExpressionPreview**
- Uses `cronstrue` a small external package for generating and localizing a verbal description of the cron expression
- Next Run is calculated locally (current vue code is using `cron-parser`  for that), and is localized using standard Javascript (`Intl`)

_Note:_ the new components are not fully localized yet - labels and messages are still English. I will handle this for all components in a (near-) future PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "@floating-ui/react-dom": "^2.1.6",
         "clsx": "^2.1.1",
+        "cronstrue": "^3.3.0",
         "next": "15.3.3",
         "next-intl": "^4.3.4",
         "react": "^19.0.0",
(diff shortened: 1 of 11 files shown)


advplyr:
I couldn't find any issues. It is a nice improvement over the Vue implementation.
Thanks!

PR #23, opened 2025-09-25
Add Linting Support

KeyboardHammer:
This pr introduces the recommended eslint settings for nextjs projects as described here: https://nextjs.org/docs/app/getting-started/installation#set-up-eslint and here: https://nextjs.org/docs/app/api-reference/config/eslint

It's much easier to add this at the start of a project, so I took some time to do it now, with the following setup:
- No changes to default/recommended rules for nextjs, typescript, and prettier
- Linting can be run manually on the entire project via `npm run lint` or `npm run lint -- --fix`
- A pre-commit hook is run automatically (via husky) on staged files, so errors in errors in changed files will block committing, but unchanged files with errors are ignored.  Warning-level issues do not block committing.
- Added eslint to the list of recommended extensions, so warnings/errors are visible in vscode with direct links to the linting docs for explanations.

There are currently ~274 errors and warnings, many of which are easy fixes.  I will open a follow up pr for many of those, but for now I only changed 1 minor issue to test the pre-commit hook

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx lint-staged
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,3 +1,3 @@
 {
-  "recommendations": ["bradlc.vscode-tailwindcss"]
+  "recommendations": ["bradlc.vscode-tailwindcss", "dbaeumer.vscode-eslint"]
 }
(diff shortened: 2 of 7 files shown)


mikiher:
Thanks! I've used your code as a basis, and also fixed all existing linter errors/warnings in PR #25 .

PR #24, opened 2025-09-26
Localize all existing components

mikiher:
This localizes all existing server and client components, by calling the `useTypeSafeTranslations` in client components and `getTypeSafeTranslation` in server components.

The type safety allows for erros to be shown for missing keys in the en-us.json file at compile time.

Also, added a prettier plugin to automatically sort translation files lexically when they are saved (no need to manually search where to add a new message anymore...)

Also, upgraded Next to the latest version to get rid of vulnerabilities.

**From now on, please use translations everywhere**


mikiher:
I think pluralization in ICU message format is more complete than what we're using currently or what i18next is providing. There are language quirks (e.g. Arabic, Russian) that aren't well covered by just adding a _plural suffix. We'll have to provide some instructions to translators, though.

mikiher (commented):
I see. I need to check what Weblate can do.

PR #26, opened 2025-09-29
Migrate BookDetailsEdit

mikiher:
This migrates the BookDetailsEdit component.

Due to the many fields in the form, this makes use of `useReducer` to maintain state.

Also have a look at the example and the use of onChange handler to enable/disable the Save button.


mikiher:
I will test BATCH_UPDATE when I implement the batch components. I wll also catch up on compoent tests at some point.

PR #28, opened 2025-10-03
Migrate Edit Genres, Tags, Narrators lists

KeyboardHammer:
Creates a common component for the interactive editable tables/lists used in:
- Edit Genres
- Edit Tags
- Narrators Page

The 'Books' Column will show/hide based on the api response passed to the component.

New to react so something may have been missed.

### DIFF
--- a/src/components/ui/EditList.tsx
+++ b/src/components/ui/EditList.tsx
@@ -2,7 +2,8 @@
 
 import { useTypeSafeTranslations } from '@/hooks/useTypeSafeTranslations'
 import { mergeClasses } from '@/lib/merge-classes'
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { TranslationKey } from '@/types/translations'
+import { Fragment, useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import Modal from '../modals/Modal'
 import Btn from './Btn'
 import IconBtn from './IconBtn'
@@ -33,9 +34,7 @@ export default function EditList({ items, onItemEditSaveClick, onItemDeleteClick
   const [isDeleting, setIsDeleting] = useState(false)
   const [hasSameName, setHasSameName] = useState(false)
   const [sameNameWithDifferentCase, setSameNameWithDifferentCase] = useState('')
-  const showNumBooks = items.some((item) => {
-    return item.numBooks && item.numBooks > 0
-  })
+  const showNumBooks = listType === 'Narrator'
 
   // Switch to edit mode for a row
   const handleEditItemClick = (clickedItem: EditListItem) => {
@@ -120,7 +119,7 @@ export default function EditList({ items, onItemEditSaveClick, onItemDeleteClick
     [editedItem, newName, handleSaveEditClick]
   )
 
-  const getListTypeEditString = useMemo(() => {
+  const listTypeEditString: TranslationKey = useMemo(() => {
     switch (listType) {
       case 'Tag':
         return 'MessageConfirmRenameTag'
@@ -132,7 +131,7 @@ export default function EditList({ items, onItemEditSaveClick, onItemDeleteClick
     }
   }, [listType])
 
-  const getListTypeDeleteString = useMemo(() => {
+  const listTypeDeleteString: TranslationKey = useMemo(() => {
     switch (listType) {
       case 'Tag':
         return 'MessageConfirmRemoveTag'
@@ -144,7 +143,7 @@ export default function EditList({ items, onItemEditSaveClick, onItemDeleteClick
     }
   }, [listType])
 
-  const getListTypeMergeString = useMemo(() => {
+  const listTypeMergeString: TranslationKey = useMemo(() => {
     switch (listType) {
       case 'Tag':
         return 'MessageConfirmRenameTagMergeNote'
@@ -156,7 +155,7 @@ export default function EditList({ items, onItemEditSaveClick, onItemDeleteClick
     }
   }, [listType])
 
-  const getListTypeWarningString = useMemo(() => {
+  const listTypeWarningString: TranslationKey = useMemo(() => {
     switch (listType) {
       case 'Tag':
         return 'MessageConfirmRenameTagWarning'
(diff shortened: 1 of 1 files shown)


src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,218 @@
...
+    delRef.current = clickedItem
+    setIsDeleting(true)
+    setIsProcessingModalOpen(true)
+  }
+
+  const handleDeleteModalClick = async () => {
+    if (!delRef.current) return
+    setIsProcessing(true)
+    await onItemDeleteClick(delRef.current)
+      .then(() => {
```
mikiher, line 57: Use of .then with await is redundant. Also there's no error handling.
```suggestion
    try {
        await onItemDeleteClick(delref.current)
    } catch(e) {
        ...
    } finally {
        ...
    }
```

src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,218 @@
...
+  const handleSaveEditClick = (clickedItem: EditListItem) => {
+    setEditedItem(clickedItem)
+    setIsDeleting(false)
+    setIsProcessingModalOpen(true)
+  }
+
+  const handleSaveModalClick = async () => {
+    setIsProcessing(true)
+    await onItemEditSaveClick(editedItem, newValue)
+      .then((updatedItemList) => {
```
mikiher, line 87: Same comment as above.

mikiher:
Also, I edited a genre and saved, then I changed it back to the original value saved, and got an error.

src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,218 @@
...
+  id: string
+  value: string
+  numBooks?: number
+}
+
+interface EditListProps {
+  items: EditListItem[]
+  onItemEditSaveClick: (item: EditListItem, newValue: string) => Promise<any>
+  onItemDeleteClick: (item: EditListItem) => Promise<any>
+  saveConfirmI18nKey: TranslationKey
```
mikiher, line 22: Don't use any. Be specific about the type.

src/app/(main)/components_catalog/examples/EditListExamples.tsx
```diff
@@ -0,0 +1,75 @@
...
+        </p>
+        <p className="mb-2">
+          <span className="font-bold">Props:</span> <code className="bg-gray-700 px-2 py-1 rounded">items</code>,{' '}
+          <code className="bg-gray-700 px-2 py-1 rounded">onItemEditSaveClick</code>,<code className="bg-gray-700 px-2 py-1 rounded">onItemDeleteClick</code>,{' '}
+          <code className="bg-gray-700 px-2 py-1 rounded">saveConfirmI18nKey</code>,<code className="bg-gray-700 px-2 py-1 rounded">deleteConfirmI18nKey</code>,{' '}
+          <code className="bg-gray-700 px-2 py-1 rounded">libraryId (optional)</code>,
+        </p>
+        <p className="mb-2 text-sm text-gray-400">Features: Automatically shows extra column for number of books for pages that support that info</p>
+      </ComponentInfo>
+      <ExamplesBlock>
```
mikiher, line 45: Please add an example with a numBooks column

mikiher:
Please use `useCallback` and `useMemo` wherever possible, to minimize re-rendering.

src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,218 @@
...
+            >
+              {item !== editedItem && (
+                <>
+                  <td className={mergeClasses('p-3')}>
+                    <a
+                      className={mergeClasses('text-sm md:text-base text-gray-100 truncate')}
+                      title={item.value}
+                      // Only narrators can be clicked on, tags/genres don't have library info attached.
+                      // href could be made an EditListItem prop that is passed in if other pages start using this
+                      href={libraryId ? `/library/${libraryId}/bookshelf??filter=narrators.${item.id}` : undefined}
```
mikiher, line 123: Double question mark.

src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,218 @@
...
+            </th>
+            {showNumBooks && <th>{i18n('LabelBooks')}</th>}
+          </tr>
+        </thead>
+        <tbody>
+          {itemList.map((item, idx) => (
+            <tr
+              key={item.id}
+              className={mergeClasses('w-full p-2 group', {
+                'bg-primary/20': idx % 2 === 0
```
mikiher, line 112: use `even:bg-primary/20` (or `odd:bg-primary/20`) instead of making javascript calculations. You can then also remove `idx`, since you don't need it.

src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,218 @@
...
+              })}
+            >
+              {item !== editedItem && (
+                <>
+                  <td className={mergeClasses('p-3')}>
```
mikiher, line 117: No need to call `mergeClasses` if there are no classes to merge.

src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,218 @@
...
+                    <a
```
mikiher, line 118: I think it doesn't make sense to show a link when there's nothing to link to (for genres and tags). Maybe pass the list type as a prop and use it here - using the existence of libraryId to decide if the `href` attribute is defined or undefined (or to decide if a link is displayed or not) is kind of ugly.

mikiher:
I am assuming that you're still working on this until you explicitly say that you're ready for review again.

KeyboardHammer:
In that case it's ready for review :)

mikiher (commented):
I also wanted to ask for keyboard navigation (Enter and tabbing) in the modal, but I realize that the Modal component doesn't provide the right tooling for that, so let's leave it as is for now.

src/components/ui/EditList.tsx
```diff
@@ -0,0 +1,297 @@
...
+        e.preventDefault()
+        if (editedItem?.name !== newName && newName.trim() !== '') {
+          handleSaveEditClick(editedItem)
+        }
+      }
+    },
+    [editedItem, newName, handleSaveEditClick]
+  )
+
+  const getListTypeEditString = useMemo(() => {
```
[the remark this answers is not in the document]
KeyboardHammer: The useTypeSafeTranslations fn will error at the spot its called when a nested fn returns an invalid value, which is how I noticed some strings did not exist before:
{64C4836F-E968-4E43-8999-F4B1621780A9}
Nevertheless it never hurts to be more explicit so I'll fix this.

src/components/ui/EditList.tsx
KeyboardHammer: Updated to use listType to show an anchor for narrators, others get a span

advplyr:
You can merge it. I'll look at it later either way

mikiher:
@KeyboardHammer, thanks a lot for submitting this PR! Looking forward to reviewing additional ones.

PR #72, opened 2026-01-18
New Library item page 

mikiher:
Sorry in advance for the copious amounts of code. It took me a lot of iterations until I was able to come up with a version I was satisfied with, and this resulted in quite a bit of code.

In a nutshell, I wanted to achieve a better and more natural item detail-editing experience than the existing one, and do away with the dense tab-based modal approach, which I think is not suitable at all for mobile devices.

What I ended up with, at a high level:
- All editing is done in the item page
- Metadata editing is done naturally in-place - if you have permission, you click on a field and edit it.
- No more Save button - edits are saved when you move away from the edit box
  - if no changes were made, no save is triggered.
- In-context navigation buttons are added to the item page:
  - Previous/next buttons switch to the previous/next item
  - Close button goes back to the containing context (library, homepage, etc...) and focuses on the item that was just closed.
- Tabbing goes through each metadata field's view and edit modes.
- Cover edit is still done in a modal, since it is too complex to edit in place.
- Match, Convert to M4b, and Embed are placed in the context menu dropdown.
  - I removed quick embed for the time being, we can decide to bring it back in some form
  - Match also brings up a modal due to the complexity of interaction.
  - I put a Quick Match button in the action buttons row.
- Empty metadata fields are not displayed
  - I added an Add Field button to the action buttons row to add a new field and edit it
- The rest of the data (Episodes, Chapters, Audio Tracks, Library Files) is standardized into accordion sections with optional action items on the accordion header.
  - e.g. the Find Episodes is moved to the Episodes accordion header
- The mobile experience is now more suitable for editing.
  - When you touch on a field, it first shows the edit button and underlines links (since there's no hovering)
  - Then you can touch the fields edit button or a link

In order for the in-context navigation to work, I needed to create a LibraryDataContext at the Library level, which is responsible for both fetching keeping all the currently loaded library entity data (items, personalized shelves, collections, series, playlists, authors). This provides the context required for next/prev/close buttons on the item page, while also centralizes data fetching and caching for the various library entities. All of these entities are now lazy-loaded on the client (_note_: the Author page is not yet part of this system, as it was added only recently by KeyboardHammer)

Please let me know what you think. 
Obviously there's some more work needed (e.g. since there's no save, we need an easy undo mechanism), but I decided to stop here since feedback is needed.


mikiher:
> > > This is working, but there is a rendering bug when collapsing the Read More where it looks like the "Read More" option is hidden then appears, causing the tables at the bottom of the page to jump positions. It is very fast so hard to tell exactly what is happening. The shortened description appears to be visible the entire time.

> > I believe I have fixed this. PTAL again.

> This is still happening, not sure what is going on here. Might be a check as to whether the description is long enough?

Sorry about this. It's really hard to notice any change on a fast computer. I added logging and there were indeed some render cycles in which the showButton state flipped from true to false and then back to true. 

I fixed this and the flicker should not be happening anymore. 

There will still be some cases in which the button will appear/disappear later than the text. The clamping calculation can only happen when the text is already laid out. I have length-based heuristic that tries to predict whether the text is clamped or not before actually calculating it, but the heuristic may provide a wrong value.

mikiher:
Closing this PR now as it's very old and is unlikely to be released anyway.

PR #85, opened 2026-03-01
Add tracks table to book page

mikiher:
This adds the audio tracks section to the book page, using `CollapsibleSection` and `DataTable`.

The Full Path button is also removed from the header of both Library Files and Audio Tracks on smaller screens, like in the old client.


advplyr:
Thanks!

PR #86, opened 2026-03-04
UserContext: Refactor UserLoginResponse into client side state

mikiher:
This change refactors `UserLoginResponse`, returned by the `getCurrentUser` api call to be kept as a client-side state in a new context `UserContext`.

The refactor was performed to make it possible for the user state to be updated on the client without requiring server refresh, including via websocket events such as `user_updated`.

**Notes**:
- Now the various fields comprising `UserLoginResponse` can be returned separately by deconstructing the `useUser` response object.
- I added a `getCurrentUser` call to `MainLayout`, passing its return value to a `LibraryProvider`
- I removed `getCurrentUser' from **most** other layouts/pages, excluding a few places where it's required for some specific server-side check (but it's OK to have it called twice in the same request since it's cached)
- I replaced `currentUser` props on main client side components with `useUser()` calls that grab the required fields from the context
  - We can reduce `currentUser` prop-drilling even further if we choose to - I didn't bother with that right now.

### DIFF
--- a/src/app/(main)/AppBar.tsx
+++ b/src/app/(main)/AppBar.tsx
@@ -3,8 +3,9 @@
 import IconBtn from '@/components/ui/IconBtn'
 import Tooltip from '@/components/ui/Tooltip'
 import { useMediaContext } from '@/contexts/MediaContext'
+import { useUser } from '@/contexts/UserContext'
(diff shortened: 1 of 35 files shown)


advplyr:
The `user_updated` socket event can contain any user. I just fixed a bug in the main client that wasn't checking the user id.

Thanks!

PR #95, opened 2026-03-22
Add listening session view

Vito0912:
I tried to partly copy the behavior from the current listening session page. The most obvious change is the delete process, which no longer replaces the header and now sits on top instead. This changes the horizontal layout, but keeps the table header in place. In the current table, it jumps around a bit when pressed.

I added the matching API calls and models. One thing I noticed in general is that they are made for general use and are not very specific, which could be a problem down the line.

Disclaimer: For the table styling, I used a bit of AI, but I refined it and kept it close to the old styling. Hope that fits

Images:
image
image
image
image
image


src/app/(main)/settings/listening-sessions/ListeningSessionsTable.tsx
```diff
@@ -0,0 +1,642 @@
...
+    setShowResumePlaybackPrompt(true)
+  }
+
+  const handleStartPlaybackAtTime = async () => {
+    if (!resumePlaybackSession) return
+
+    setStartingPlayback(true)
+
+    try {
+      const libraryItem = await getExpandedLibraryItemAction(resumePlaybackSession.libraryItemId)
```
advplyr: We need to show the failed to load data toast if the libraryItem is not found. At some point it would be great to have the sessions API endpoint check whether the item still exists so we can know if the timestamp should be clickable. For now though we just show the failed to load data toast.

PR #100, opened 2026-04-08
Modal prev/next navigation and MatchModal integration

mikiher:
This adds prev/next navigation buttons to `MatchModal`, in a way that can be reused for other modals (e.g. Edit, AuthorEdit) in the future.

- In the case of `LibraryClient,` the matched card's shelf is used as navigation context.
- In the case of `BookshelfClient`, the surrounding contiguous non-null items block is used as navigation context.

The navigation context is frozen when the modal is opened, so if anything changes underneath (items updated/added/reomved), the context remains the same and navigation is predictable.

This also removes Match from podcast episode cards, as discussed on Discord.

### DIFF
--- a/src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx
+++ b/src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx
@@ -335,6 +335,7 @@ export default function BookshelfClient({ entityType }: BookshelfClientProps) {
                     )
                   }
 
+                  const entityIndex = startIndex + k
                   return (
                     <config.CardComponent
                       key={`card-wrapper-${item.id}`}
@@ -347,6 +348,8 @@ export default function BookshelfClient({ entityType }: BookshelfClientProps) {
                       orderBy={orderBy}
                       seriesSortBy={seriesSortBy}
                       bookProgressMap={bookProgressMap}
+                      shelfEntities={entityType === 'items' ? items : undefined}
+                      entityIndex={entityType === 'items' ? entityIndex : undefined}
                     />
                   )
                 })}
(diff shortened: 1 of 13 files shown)


src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx
```diff
@@ -347,6 +348,8 @@ export default function BookshelfClient({ entityType }: BookshelfClientProps) {
                       orderBy={orderBy}
                       seriesSortBy={seriesSortBy}
                       bookProgressMap={bookProgressMap}
+                      shelfEntities={entityType === 'items' ? items : undefined}
```
Comment (advplyr) [inline src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx:351]: Do you think there is additional overhead by passing the entire array of item objects into each card component? I'm not sure how react adds listeners to props. If it does we could limit it to pass the entity id's

Comment (mikiher) [reply src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx:351]: I'm not sure I follow. This is just a reference. When a match modal is opened we call `getEntityNavigationContext` to get a contiguous non-null block of ids surrounding the matched entity using it, but until then, it's just an array reference. I don't want to create a long array of entity ids for no reason.

Comment (advplyr) [reply src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx:351]: I dug into this a bit and I don't think it adds any overhead. It may be cleaner still to only pass the IDs but no other benefit

advplyr:
works great, thanks!

src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx
Comment (mikiher) [reply src/app/(main)/library/[library]/[entityType]/BookshelfClient.tsx:351]: If we had the ids in an array for some other purpose, I'd agree, but I don't want to create it just for the benefit of passing a "cleaner" prop, becuase it's a waste of memory. 

A wiser use of memory would be to create a more frugal minified version of libraryItem, so a non-sparse items array  is less memory intensive.

PR #102, opened 2026-04-11
progress updates, progress for podcast, episode title in miniplayer

Vito0912:
Same as title

Currently the title jumps around for podcast episodes. Maybe there is a better way to handle that for now. Didn't want to change too much


advplyr:
I was going to update the `bookProgressMap` to be a `mediaItemProgressMap` but it became too big of a change so I'll do that in a separate PR.

Thanks!

PR #104, opened 2026-04-12
Integrate edit modal in media card

mikiher:
This integrates `LibraryItemEditModal` in media cards on the homepage and library (opened using the edit button on the overlay).

Some refactoring was needed to make `MatchModal` and `LibraryItemEditModal` use the same item fetching and navigation capabilities: `LibraryItemModal` was added as parent class in both modals, providing the mentioned capabilities.

The only differences between the edit modal when opened from a media card and from the item page are:
- Expanded item fetching
- Prev/Next buttons
- When opened from a media card, the height of the modal content is fixed, so that it doesn'r change when navigating to prev/next. This means that the book/podcast details may overflow if they're long, and a vertical scrollbar might be shown (this is similar to behavior of the current edit details tab).

### DIFF
--- a/src/components/modals/LibraryItemEditModal.tsx
+++ b/src/components/modals/LibraryItemEditModal.tsx
@@ -1,42 +1,111 @@
 'use client'
 
 import { updateLibraryItemMediaAction } from '@/app/actions/mediaActions'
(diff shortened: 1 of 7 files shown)


advplyr:
Works great, thanks!

PR #108, opened 2026-05-03
Collection Page

mikiher:
This implements the collection page in a bookshelf style instead of the existing table style. I wanted to get your opinions on it before continuing.

I wanted the books in a collection to resemble media cards as much as possible for consistency with the other bookshelfs, while adding the minimal controls required for collection management and reordering.

*This is currently only for review purposes. Please don't merge.*

### Notabale changes
- Visually, the main differences in a collection book card are:
  - The drag handle
  - The Mark as Finished/Not Finished button
- I didn't implement a "Remove from Collection" button yet. Not sure where to put it yet (maybe next/instead of the "Mark as Finished" button?)
- I initially implemented it using the existing FormKit, but it is has issues on mobile which I was not able to resolve. 
  - Switching to dnd-kit resolved all the mobile issues - dragging, scrolling and animations seem to work very smoothly.
  - dnd-kit seems more mature and less buggy, and the drag animations are superior
  - It has built in keyboard support!
  - If you want to see the FormKit based version, rollback the last commit
- Like the existing Vue implementation, there's no virtualization. I'm quite sure it can be baked in later, but I think it requires a change in the API.

Please let me know what you think.


advplyr:
I think it might work with some ui changes:
- The drag handle should only show on hover. Probably placed at the top middle of the card.
- The mark as finished and remove from collection should be context menu items. I don't think we need another overlay for those.
- We should ignore the bookshelf view setting. Currently if you enable bookshelf view the title disappears. I don't think we should support bookshelf view on pages that will have top section.
- The floating "Collection Items" header looks out of place but I don't have a good suggestion yet. Maybe just dropping the background color and x padding.

mikiher:
> * The drag handle should only show on hover. Probably placed at the top middle of the card.

Done. I still think it's more natural on the side, though.
Also, on mobile it has to appear always since there's no hover.

> * The mark as finished and remove from collection should be context menu items. I don't think we need another overlay for those.

Done.

> * We should ignore the bookshelf view setting. Currently if you enable bookshelf view the title disappears. I don't think we should support bookshelf view on pages that will have top section.

Done.

> * The floating "Collection Items" header looks out of place but I don't have a good suggestion yet. Maybe just dropping the background color and x padding.

I removed the header and moved the info (number of books and total duration) to the Toolbar (similar to other bookshelf pages). 

PTAL again. Let me know what you think, but again, please don't merge yet.

advplyr:
This is looking good. I think the handle at the top center looks better than the side.
There is an issue where the handle stays visible after dragging and releasing.

I'm not sure how we should handle mobile. Sorting on mobile isn't a great experience in general. Having the handle permanently overlaid on the cover wouldn't be popular.
We could start with hiding the handle on mobile and make it a desktop feature.

mikiher:
> This is looking good. I think the handle at the top center looks better than the side. There is an issue where the handle stays visible after dragging and releasing.

Fixed.
 
> I'm not sure how we should handle mobile. Sorting on mobile isn't a great experience in general. Having the handle permanently overlaid on the cover wouldn't be popular. We could start with hiding the handle on mobile and make it a desktop feature.

Now mobile reordering is off by default. I implemented a Reorder Books button that shows the drag handles and allows reoerering. You can turn it off when you're done. How does that look? I think that with dnd-kit the mobile reordering experience is quite good, even when combined with scrolling.

Also I switched drag handle to the Material Symbols `drag_handle`

advplyr:
I think that works. The handle is now permanently showing for desktop

mikiher:
Sorry, I believe I inadvertantly broke the tailwind "no dynamic class constuction" rule, which caused the issue on desktop. Is it fixed for you now?

advplyr:
I think that the drag handle should look like the other overlay icons. Is it necessary to have CollectionBookCardShell and the drag handle be separate from all the other overlay icons?
I was going to go in and tweak some of the CSS but it's difficult to work with.

Right now if you hover over the handle the overlay disappears. The drag handle also has its own background and is not using the same padding and color as the other icons. If it is difficult to see without the separate background, then we can darken the hover overlay for the card.

I was testing on mobile also and I'm not sure how it's supposed to work. If you tap on a card you navigate to the book page. Until we find a way to work around the hover overlay I think we should leave re-ordering out of mobile.

mikiher:
> I think that the drag handle should look like the other overlay icons. Is it necessary to have CollectionBookCardShell and the drag handle be separate from all the other overlay icons? I was going to go in and tweak some of the CSS but it's difficult to work with.

I think the drag handle isn't just a standard overlay icon. It does something fundamnetally different than other standard overlay icons. I think it should be large enough to be a touch target on mobile, but I'm fine with whatever styling decisions you make. I'm also not married to the separate components. Tell me how you want it and I'll implement.

> Right now if you hover over the handle the overlay disappears. The drag handle also has its own background and is not using the same padding and color as the other icons. If it is difficult to see without the separate background, then we can darken the hover overlay for the card.

I am fine with whatever style decision you make, and also with making it part of the standard overlay. The things I think are important:
- I think it should be large enough to be a touch target on mobile.
- I think it's important that it's visible on all covers (perhaps the overlay is dark enough if it's added as part of it)
- It should also be visible in mobile reorder mode. There's no hover overlay on mobile, so either we display an overlay in reorder mode or keep the drag handle semi-opaque like it is currently.

> I was testing on mobile also and I'm not sure how it's supposed to work. If you tap on a card you navigate to the book page. Until we find a way to work around the hover overlay I think we should leave re-ordering out of mobile.

You're supposed to press the "Reorder Books" button (next to the Collection Edit button), and then drag handles become visible until you press "Done Reordering". There's no hover overlay on mobile.

advplyr:
> You're supposed to press the "Reorder Books" button (next to the Collection Edit button), and then drag handles become visible until you press "Done Reordering". There's no hover overlay on mobile.

Right now when you press "Reorder Books" nothing happens besides it becomes available when hovering over the card, but there is no hovering on mobile.

I'm perfectly fine with not supporting this on mobile. If we do support it then I guess we show the full card overlay and only have the handle visible. I think we make the overlay bg slightly darker and use that instead of having the handle use its own background.

We should move the handle to be with the other overlay icons and it should use the same styling as the other icons. We can make it larger if we need to but it should be consistent with the others.

mikiher:
> Right now when you press "Reorder Books" nothing happens besides it becomes available when hovering over the card, but there is no hovering on mobile.

Hmm, this is working perfectly for me:
- The Reorder Books button only shows up on mobile.
- When you press Reorder Books (on mobile), all the drag handles appear. 
- When you press it again (Done Reordering), all the handles disappear.

Which browser are you using for testing on mobile?

advplyr:
Ok that is working actually. Not sure, maybe I missed it. The full bg overlay would help to see that it is in drag mode

mikiher:
OK, I've made the requested changes, and I think I'm happy now with the code.

- The drag handle is now displayed with all the other overlay buttons, and is based on the same overlay button component.
  - I made it slightly larger than the other components - you can change it in `DraggableMediaOverlayIconBtn`
- On mobile (no-hover surfaces, to be exact), when switching to reorder books mode, the drag handle is display on the overlay.
- I have renamed many of the components (+ additional small changes) in preparation to making the Playlist page use the same reorder mechanism.

The code is ready for review from my perspective.

After you approve this, I will switch SortableList from formkit to dnd-kit and remove formkit, in a separate PR.

Also, I'm aware that the Play All button is not implemented. I'll also add this in a separate PR.

advplyr:
I don't think we can build the playlists page the same way because of podcast episodes.
Since every podcast episode has the same cover it wouldn't look right to have a full page of cards all with the same cover.

This is working well in my testing. Thanks!

mikiher:
What about book playlists?

advplyr:
hmmm do you think it is better to use a single consistent page layout or the layout be different based on the media type?

mikiher:
I think book collections and book playlists should have (roughly) the same experience. We already talked sometime recently that the difference between them was not so obvious (even to me). We shouldn't artificially differentiate book collections and playlists just because podcast playlists are a somewhat different beast (in which the grid UI makes less sense)

advplyr:
I started a discussion on Discord and I think the best option is to build both views and have a toggle button to switch between the views, then book media type would default to the cover view and podcast media type would default to the list view.
There have been requests to have both views

Vito0912:
Link for reference: https://discord.com/channels/942908292873723984/954760207131615264/1504824485352837222

advplyr:
I found a bug that is a bit strange. I wonder if you can reproduce this.
1. Start with a page width > 1200px
2. Resize the page to < 1000px
3. Drag a book off the right side of the page

observe side rail is pushed off the left side of the screen and empty bg is shown on the right.
screenshot is full width
image

mikiher:
> I started a discussion on Discord and I think the best option is to build both views and have a toggle button to switch between the views, then book media type would default to the cover view and podcast media type would default to the list view.

I think that is a good solution. I will implement the list view and the switch button with default based on media type.
I'm assuming this is only for the collection and playlist pages, right? A list view for the library page is more complex.

> I found a bug that is a bit strange. 

I think I encountered something similar when re-implementing `SortableList` with dnd-kit, and resolved it. I'll look into this.

advplyr:
> I'm assuming this is only for the collection and playlist pages, right? A list view for the library page is more complex.

Yes, but eventually we will need to make collection and playlist items paginated. Just don't need to do that before release

PR #111, opened 2026-05-14
Library Stats Page

KeyboardHammer:
Implements the Library Stats page.

I did change some of the numeric formatting and labels to use the common formatting functions, so it's not a 100% match with the vue ui.  I also did not implement the percentage bars on the authors, longest, and largest sections.  I'm not a huge fan of them in the vue ui and find them a little confusing, but I don't care much either way and can re add them if you still want them.

I did test on mobile, and made the sections more responsive than before.

{F0AD4270-5C4A-4AAE-AA2D-7AEFBC6381D5}


advplyr:
I like a bit more spacing. I agree the bars never made much sense. Thanks!

PR #113, opened 2026-05-15
Remove formkit dependencies

mikiher:
This re-implements `SortableList` using dnd-kit, and removes the dependency on formkit.

### Notable changes:
- `LibrariesList` and `LibraryScannerTab` now use the modified `SortableList`
- Drag handle classes are refactored out of `DraggableMediaOverlayIconBtn` to be used by all drag handles
- Drag handles are using the `drag_handle` icon across the code for consistency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
     "@dnd-kit/sortable": "^10.0.0",
     "@dnd-kit/utilities": "^3.2.2",
     "@floating-ui/react-dom": "^2.1.6",
-    "@formkit/drag-and-drop": "^0.5.3",
     "clsx": "^2.1.1",
     "cronstrue": "^3.3.0",
     "date-fns": "^4.1.0",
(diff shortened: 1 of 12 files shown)


mikiher:
The last commit fixes the issue reported [here](https://github.com/audiobookshelf/audiobookshelf-client-react/pull/108#issuecomment-4459940960). 

The problem had to do with `Tooltip` instances inside `MediaCardDetailView`, which caused overflow issues when rendered inside a drag overlay. To fix this, I changed the Tooltip to use the `fixed` floating-ui strategy (instead of `absolute`) and use a portal by default - both were recommended by the AI as better default options for Tooltip to avoid such layout and overflow bugs. The tooltip behavior remains functionally the same.

advplyr:
There is an issue with dragging inside a scrollable container. If you drag an item down the scroll container expands.

image

mikiher:
> There is an issue with dragging inside a scrollable container. If you drag an item down the scroll container expands.

Ok, will look into this.

mikiher:
- Fixed vertical overflow issue
- Fixed issue with dragging pointer icon disappering after drag start

mikiher:
Also, in the library scanner tab, I disabled dragging for disabled sources.

advplyr:
This is working great, thanks!

PR #114, opened 2026-06-04
Add support for cross-origin dev requests

mikiher:
This adds support for allowing access to cross-origin dev resources, which is blocked by default in Next.js 16.

To support this, I sent audiobookshelf [PR #5291](https://github.com/advplyr/audiobookshelf/pull/5291) which sets the `ALLOWED_DEV_ORIGINS` env variable from a `dev.js` option. This PR reads this variable and sets the `allowedDevOrigins` in `next.config.ts`

I updated the README file with proper instructions.


advplyr:
Works for me!

PR #116, opened 2026-06-09
Fix keyboard navigation for SortableBookshelf

mikiher:
This makes the required changes to make keyboard navigation work correctly in `SortableBookshelf`.

The follwing behavior is implemented:
- Tab moves between cards (only one tab stop per card)
- Arrow keys take the bookshelf into reorder mode and also move the card
- In reorder mode:
  - Enter/Tab accepts new card position and goes out of reorder mode
  - Escape reverts to the original card position and goes out of reorder mode
  - Arrow keys move the card


advplyr:
Oh nice, that works great. Thanks!

PR #117, opened 2026-06-10
Add Collection list view and bookshelf/list toggle button

mikiher:
This adds a list view that shows the collection books in a table similar to the current Vue implementation, and adds a toggle button to the collection header to switch between list and bookshelf views. A similar toggle will be added to the Playlist page when it's implemented.

*Note: Please do not merge yet. This is still for functionality review purposes only.*

### Notable list view changes:
- On desktop list view, the righthand-side list row buttons are all invisible by default and appear only on hover.
- On mobile list view, the drag handle is invisible by default, and is shown when the reorder mode button is clicked, similar to bookshelf behavior. The righthand-side row buttons are not displayed.

I have a couple of questions before I proceed:
1. Keyboard navigation in list view is currently not similar to bookshelf view - there are multiple tab stops per list row. What's the prefered keyboard nav behavior for list view?
2. The Mobile app has a play button on each collection book row. Should we also add one on mobile list view?
3. On mobile, there are currently a few touch targets like on desktop (book title takes you to book page, author takes you to author page). I think we should make each row a single touch target on mobile (which takes you to the book page), right?


advplyr:
Switching between views and the drag and drop is working great for me.

1. For list view tab stops I think we keep them how they are now. When the drag handle is selected you can re-order by pressing arrow up or down.
2. Yeah a play button would be nice for mobile to show on the right
3. Single touch target to the book page makes the most sense to me

mikiher:
OK, please take a look now:
- I've left the tab stops as requested, with arrow up/down reordering when drag handle is focused
- Fixed focus outline clipping issues when tabbing into row links (title, series, and authors)
- Made the whole row a single touch target on mobile
- Added the play button at the end of the row on mobile
- Made the list view responsive to card size increase/decrease
- Fixed and issue with media progress retrieval in list view

This is ready for review and merge now, from my perspective.

advplyr:
This is working great for me. I couldn't find any issues. Thanks!

PR #120, opened 2026-06-20
Episode edit and match modals

mikiher:
This implements the episode edit and match modals.

### Notable changes:
- Supports context navigation in all episode card/list-row contexts:
  - Playlist
  - Homepage episode shelves
  - Episode table in podcast page

### Other changes and fixes:
- ViewEpisodeModal is opened:
  - when an episode card is clicked 
  - when episode title is clicked in playlist list view
  - when episode is clicked in playlist list view on mobile
- Podcast title in ViewEdpisodeModal now links to the podcast page
  - on mobile, the link is always underlined
  - on desktop it's underlined on hover
- Fixed bug where Escape didn't close a modal if one of its descendants was not focused.
- Added episode badge to episode cards in playlist bookshelf view
- Fixed bug where episode card context menu stayed open when the card was clicked
- Added context navigation to ViewEpisodeModal (in all contexts)

### DIFF
--- a/src/components/widgets/EpisodeDetailsEdit.tsx
+++ b/src/components/widgets/EpisodeDetailsEdit.tsx
@@ -1,0 +1,213 @@
+'use client'
+
+import Dropdown, { DropdownItem } from '@/components/ui/Dropdown'
+import SlateEditor from '@/components/ui/SlateEditor'
+import TextareaInput from '@/components/ui/TextareaInput'
+import TextInput from '@/components/ui/TextInput'
+import { useTypeSafeTranslations } from '@/hooks/useTypeSafeTranslations'
+import { formatJsDate } from '@/lib/datefns'
+import type { PodcastEpisode, UpdatePodcastEpisodePayload } from '@/types/api'
+import { useCallback, useEffect, useImperativeHandle, useMemo, useReducer, useRef, useState } from 'react'
+
+type EpisodeDetails = {
+  season: string
+  episode: string
+  episodeType: string
+  title: string
+  subtitle: string
+  description: string
+  pubDate: string | null
+  publishedAt: number | null
+}
+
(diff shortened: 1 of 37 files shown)


src/components/widgets/EpisodeDetailsEdit.tsx
```diff
@@ -0,0 +1,213 @@
...
+        <div className="w-2/5 p-1">
+          <TextInput
+            ref={pubDateInputRef}
+            type="datetime-local"
+            value={pubDateInput}
+            onChange={handlePubDateChange}
+            label={t('LabelPubDate')}
+            error={pubDateInvalid ? t('ToastDateTimeInvalidOrIncomplete') : undefined}
+          />
+        </div>
```
advplyr: The date input error doesn't clear when selecting a date from the browser calendar.
After making it invalid, select a date from the calendar and the error remains.

I'm also not a fan of the error message pushing everything else down. I think we could leave off the error message and only have the error border
Image

advplyr:
This is an existing bug. The view episode modal is too small when there is no description. It should be the same width when navigating next/prev episodes

image

src/components/widgets/EpisodeDetailsEdit.tsx
mikiher: Fixed.

mikiher:
> This is an existing bug. The view episode modal is too small when there is no description. It should be the same width when navigating next/prev episodes

Fixed. The modal now has a fixed width on sm and above.

mikiher:
Also fixed a broken test, and some mobile layout issues in the edit episode details modal.

mikiher:
Also added a few EpisodeTable fixes:
- Improved mobile layout on narrow screens
- Decreased count badge font when badge is long to prevent wrapping
- Made all EpisodeTableToolbar controls have the same height

advplyr:
Working great, thanks!

PR #121, opened 2026-06-22
Add cover edit modal and integrate into library item page and media cards

mikiher:
This fixes issues in the `CoverEdit`, creates `CoverEditModal`, and integrates into the library item page and library item media cards.

### Notable Changes
- Fixed various `CoverEdit` layout and controls issues, and simplified the layout
- Added `CoverEditModal` with `LibraryItemModal` as parent for navigation context and expanded item fetching
- Fixed faulty `Cancel` search button handler
- Added "Edit Cover" context menu items to the library item page and media cards.


advplyr:
I'm getting varying results on if the preview cover image changes. When editing the cover image on the home page, sometimes the preview image changes and sometimes it doesn't.
Doing an upload on the home page doesn't seem to ever update the preview. When changing the cover image with an external URL it works the first time, but usually not the second time.

I can't pinpoint an exact pattern. 

A separate issue is that if the access token expires the preview image shown in local images doesn't load and shows 401 unauthorized. I have my access token expiry set to 10 seconds so this was happening frequently in my testing.

image


The upload preview has a UI issue:

image

When trying to upload a larger image this error gets thrown:

image
This was happening on all the images 843KB and above. The next size I tested was ~500KB and that worked fine.
In the Vuejs app I was able to upload all of my test images up to 4MB.

mikiher:
> I'm getting varying results on if the preview cover image changes. When editing the cover image on the home page, sometimes the preview image changes and sometimes it doesn't. Doing an upload on the home page doesn't seem to ever update the preview. When changing the cover image with an external URL it works the first time, but usually not the second time.

Fixed. Put socket listeners in place for item updates where they were missing.

> A separate issue is that if the access token expires the preview image shown in local images doesn't load and shows 401 unauthorized. I have my access token expiry set to 10 seconds so this was happening frequently in my testing.

Fixed. added refresh logic to both the item file route, and `PreviewCover`

> The upload preview has a UI issue:

Fixed. Now PreviewCoverModal is used for both search results and upload

> When trying to upload a larger image this error gets thrown:

Fixed, but depends on [server PR #5325](https://github.com/advplyr/audiobookshelf/pull/5325).

The first problem was that server actions in next.js have limitiations on the size of data passed.
Then after fixing this by switching to a dedicated route for file upload, the second problem was that the abs server consumed the multipart body before next.js hada chance to do so - this is what PR#5325 fixes.

mikiher:
All the issues reported should now be fixed. PTAL again.

mikiher:
I updated cover, backup, and item uploads to all use the same method, as discussed on Discord:
- Upload is handled through internal-api routes, using stream piping (which avoids large memory allocation)
- Route handlers share most code
- Client side code differs slightly, with backup and item uploading continuing to use `XmlHttpRequest` for upload progress updates, while cover upload using standard `fetch`.
- Client side code does not need to retrieve or pass an access token anymore

I tested all the relevant paths and they seem to work correctly on large files, without any visible performance issues.

PTAL again.

mikiher:
Also, it looks like when you switched to Next16, you mistakenly removed next/typescript lint rules. I re-introduced them (and added a couple of ignore directories), and fixed all existing lint errors. Let's keep the linter happy!

src/components/covers/PreviewCover.tsx
```diff
@@ -81,28 +95,37 @@ export default function PreviewCover({
...
-    (err: React.SyntheticEvent<HTMLImageElement, Event>) => {
+    async (err: React.SyntheticEvent<HTMLImageElement, Event>) => {
+      if (!authRetryAttemptedRef.current && isAuthenticatedImageUrl(src)) {
+        authRetryAttemptedRef.current = true
+        const refreshed = await silentRefreshSession()
+        if (refreshed) {
+          resetImageState(withImageRetryParam(src))
+          return
+        }
+      }
```
advplyr: I can't think of why this is necessary. The token refresh is happening in the `/internal-api/` file route now so if there was a necessary refresh it would happen there

advplyr:
Are uploads working for you? I get a 500 error when trying cover uploads and library item uploads.

```
 POST /library/bc044912-f6a9-46dd-9c38-ddfdd96a12eb 200 in 35ms (next.js: 5ms, proxy.ts: 4ms, application-code: 26ms)
  └─ ƒ getMetadataProvidersAction() in 21ms src/app/actions/providerActions.ts
⨯ TypeError: Response body object should not be disturbed or locked
    at ignore-listed frames {
  page: '/internal-api/items/a9c8d8cf-e706-47e9-8ed4-36120985b16d/cover'
}
 POST /internal-api/items/a9c8d8cf-e706-47e9-8ed4-36120985b16d/cover 500 in 689ms (next.js: 616ms, application-code: 74ms)
 ```

src/components/covers/PreviewCover.tsx
mikiher: Remvoed the unnecessary refresh logic.

mikiher:
> Are uploads working for you? I get a 500 error when trying cover uploads and library item uploads.

Did you apply the server fix? This PR depends on it to work.

advplyr:
Nope, forgot

advplyr:
Thanks!

PR #122, opened 2026-06-27
Implement search results page

mikiher:
This implements the search page (reached by pressing enter when typing a query in the global search box).

### Notable change from Vue implementation
- Contains tag and genre results
  - Tag and genre cards derived from the old narrator card and now share the same component (`MetadataFilterCard`). Used same icons as in global search dropdown
  - Tried a more complex tag/genre card with group cover containing the most recent books, but decided against it due to visible lag - properly showing group covers for tags and genres requires a server change which is not planned right now.
  - Also decided against introducing collection and playlist results (they too require an api call which I didn't want to add right now).

### DIFF
--- a/src/app/(main)/library/[library]/Toolbar.tsx
+++ b/src/app/(main)/library/[library]/Toolbar.tsx
@@ -48,8 +48,7 @@ export default function Toolbar() {
     onContextMenuAction?.(action)
   }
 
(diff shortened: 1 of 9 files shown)


advplyr:
Good call on keeping the UI design simple for now. This page came out great, thanks!

PR #124, opened 2026-06-28
Fix broken Dropdown submenu clicks

### LINKED ISSUE #123: Dropdown submenu selections ignored after capture-phase click-outside change
## Summary

Selecting an item from a `Dropdown` submenu (e.g. library filter Genre, Series, Progress) does nothing: the menu closes but the value does not change and the bookshelf is not filtered.

Top-level items without submenus (All, Issues, Abridged, etc.) still work.

## Steps to reproduce

1. Open a library bookshelf page with the filter dropdown (e.g. `/library/{id}/books`)
2. Open the filter dropdown
3. Hover or navigate into a submenu (Genre, Series, Progress, Missing, etc.)
4. Click a submenu item

**Expected:** Filter applies, dropdown label updates, bookshelf reloads with filtered results.

**Actual:** Menu closes (or submenu click is ignored) and filter stays unchanged.

## Root cause

Introduced in `09a7bcd` (*Fix dropdowns using portal inside modals not closing when click outside*), which switched `useClickOutside` to the **capture** phase for `Dropdown` and related components.

`DropdownMenu` submenus are rendered in a **separate portal** on `document.body` (since `fceb092`), so they are not DOM children of the main menu `menuRef`. With capture-phase outside-click detection:

1. Document listener runs before the submenu item\'s `onClick`
2. Click is treated as outside `menuRef`
3. `closeMenu()` runs and unmounts the menu
4. Submenu selection handler never fires

This did not occur with bubble-phase listening because the item `onClick` ran before the outside-click handler.

mikiher:
This fixes #123

### Fix

Treat portaled submenu fragments as inside the dropdown for click-outside purposes:

- Modify `useClickOutside` to accept an additional "inside" predicate function as argument, `isAdditionalInside`
- Tag submenu roots with `data-dropdown-id`
- Pass `isInsideSubmenu` to `useClickOutside` from `Dropdown` (creaed using `createAdditionalInsideCheck`)

Inlining submenus under `menuRef` (like `ContextMenuDropdown`) was considered but rejected because portaled submenus avoid clipping/positioning issues with the scrollable main menu (`overflow-y-auto`).

### Affected components

- `Dropdown` with submenus (library filter/sort, player settings, etc.)
- Not affected: `ContextMenuDropdown` (submenus stay inside menu ref), `InputDropdown`, `GlobalSearchInput` (menu ref covers portaled root)

### DIFF
--- a/cypress/tests/components/ui/Dropdown.cy.tsx
+++ b/cypress/tests/components/ui/Dropdown.cy.tsx
@@ -461,6 +461,15 @@ describe('<Dropdown />', () => {
       cy.get('@onChangeSpy').should('have.been.calledWith', 'sub1')
     })
 
+    it('selects subitem on click with usePortal', () => {
+      const onChangeSpy = cy.spy().as('onChangeSpy')
+      cy.mount(<Dropdown items={itemsWithSubmenus} onChange={onChangeSpy} usePortal />)
+      cy.get('button').click()
+      cy.get('[role="listbox"] > li').eq(1).trigger('mouseover')
+      cy.get('[role="menu"] li').first().click({ force: true })
+      cy.get('@onChangeSpy').should('have.been.calledWith', 'sub1')
+    })
+
     it('closes menu after subitem selection', () => {
       cy.mount(<Dropdown items={itemsWithSubmenus} />)
       cy.get('button').click()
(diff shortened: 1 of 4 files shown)


advplyr:
Thanks!

PR #125, opened 2026-06-28
Implement Edit Chapters page

mikiher:
This migrates the Edit Chapters page.

### Notable changes from Vue implementation
- Added animations to Insert/Remove chapter so that the insertion/removal is more visible
  - without animation the change is sometime tricky to observe
- Slightly increased the font size to `text-sm` on duration and title, to make them more legible
  - "Show seconds" text input remains `text-xs`
- You can now use Save/Reset after "Remove All"
  - It makes "Remove All" behavior consistent with the rest of the edit actions.
- "Edit Chapters" appears as context menu item in book cards and the book page
  - There is no "View Chapters" modal similar to Chapters tab
  - Chapters can be viewed on the book page like before, include "Edit Chapters" on the section header
    - Also changed header button to "Add Chapters" when there are no chapters

*Note: I have not worked on mobile layout for this page yet. Due to the density of this page, this may require significant changes, and is out of scope of this PR.*


src/components/widgets/ChaptersEmptyState.tsx
```diff
@@ -0,0 +1,36 @@
+'use client'
```
advplyr: I don't think this needs to be a separate component. We could show the ChaptersTable component even if it is empty and add an empty state there.

I'm not sure if defaultExpanded was intentional here or not. I don't think it should be start expanded since it looks strange.

Image

mikiher: Sure, I'll fold the separate component into ChaptersTable.

I don't thing defaultExpanded was intended. I'll fix.

mikiher: Fixed. PTAL again

src/lib/chapters/chapterEditorUtils.ts
```diff
@@ -0,0 +1,382 @@
...
+  mediaDuration: number,
+  messages: ChapterValidationMessages
+): { chapters: EditableChapter[]; hasChanges: boolean } {
+  let previousStart = 0
+  let hasChanges = chapters.length !== existingChapters.length
+  const updated = chapters.map((chapter, i) => {
+    const next = { ...chapter }
+    next.id = i
+    next.start = Number(next.start)
+    next.title = (next.title || '').trim()
```
advplyr: This `trim()` is getting called on every key press when typing a chapter title so you can't use the space key when typing.
This may be part of the sluggish page speed

advplyr:
I'm not sure about the remove all chapters logic. In Vue it is confusing because a "Save" button is shown after removing, but with this implementation a confirmation modal is shown before actually saving.

Remove all also doesn't show the placeholder.

image

It seems like there is a tradeoff with both implementations

src/lib/chapters/chapterEditorUtils.ts
mikiher: OK, I implemented some major performance rehauls to fix page sluggishness:
- Title editing now uses local draft state in ChapterRow
  - Validation and parent state updates are deferred to blur and save instead of running on every character
  - title trim is performed on blur and save only
  - hasChanges is computed on blur and save only
- Removed row add/delete animations
  - This requires a more lightweight animation than a `TransitionGroup` per `ChapterRow`, and will be implemented in a separate PR
- Implemented lazy modals to reduce page load time
- Implemented lazy tooltips instead of the stamdard ones
  - there were a number of tooltips per chapter row which caused the page to be very bloated

The page now loads quite quickly, and I don't notice any lags anymore while editing and moving between fields.

mikiher:
> I'm not sure about the remove all chapters logic. In Vue it is confusing because a "Save" button is shown after removing, but with this implementation a confirmation modal is shown before actually saving.
> 
> Remove all also doesn't show the placeholder.
> 
> It seems like there is a tradeoff with both implementations

I changed the Remova All behavior as follows:

- No confirmation dialog when clicking Remove All (Reset still confirms, as the undo path)
- Clears chapters locally and shows the empty placeholder row instead of a blank list
- Save treats the clear-all placeholder as an empty payload (isClearAllChaptersState)
- Remove All button visibility is based on on-screen chapters (hasNonPlaceholderChapters), so it appears for staged edits like “Set Chapters From Tracks” even when nothing is saved yet

Let me know if the new Remove All behavior is acceptable now. I think it is now consistent with all other edit actions on the page, which can be reset or saved.

advplyr:
The remove all logic is better and the page feels snappy.

When editing a book using the edit icon on the page it sets the duration in the top right to 0:00. The edit modal is returning a non-expanded library item which doesn't include the duration.
image


Also, when you have a long audio file name clicking on it expands. Not sure if that was intentional?
image

mikiher:
> The remove all logic is better and the page feels snappy.
> 
> When editing a book using the edit icon on the page it sets the duration in the top right to 0:00. The edit modal is returning a non-expanded library item which doesn't include the duration. 

Wow, that was quite a nasty bug. It also affected the library item page. Fixed by waiting for the socket `item_updaed` event.

> Also, when you have a long audio file name clicking on it expands. Not sure if that was intentional? 

Not intentional. Fixed.

Also fixed:
- A layout bug causing the chapter preview timer to overlap with the tracks panel
- Standardize on `xl` for displaying two columns (instead of custom width)
- Now using new `LazyTruncatingTooltipText` component for displaying on-demand tooltip for truncated filenames
  - generally I'd recommended using `LazyTooltip` and `LazyTruncatingTooltipText` for *recurring* tooltips (e.g. inside lists/tables), because the tooltip is an expensive DOM object. I'll do a pass on the code later to see if there are other places we should switch to lazy tooltips

advplyr:
Thanks!

PR #127, opened 2026-06-29
Implement tools page

mikiher:
This implements the tools page. Similar to the chapters page, the embed and encode m4b tools (sharing the same page) are wired into the book page and media card through context menu actions.

Based on our discussion on discord I decided not to automatically re-scan after m4b encoding, since this requires careful server scrutiny and potential fixes.

Also, similarly to the chapter edit PR, I did not look at mobile layout, since the tools page is dense and likely requires some redesign on mobile, which is out of scope for this migration PR.


advplyr:
> Also, similarly to the chapter edit PR, I did not look at mobile layout, since the tools page is dense and likely requires some redesign on mobile, which is out of scope for this migration PR.

Agree and this will apply to the audio track manager page too.

Thanks!

PR #128, opened 2026-06-29
Fix RTL issues in ereader TOC drawer

mikiher:
This fixes a couple of RTL issues in the ereader TOC drawer, so that Hebrew/Arabic chapter titles and search results are aligned properly to the right.

- It adds `dir="auto"` to the TOC/result buttons
- It applies `w-full` style to the buttons so the right alignment shows even when the interface language is not RTL.

screenshots:
TOC before:
image

TOC after:
image

Search results before:
image

Search results after:
image


advplyr:
Thanks!

PR #129, opened 2026-07-03
Manage tracks page

mikiher:
This migrates the Manage Tracks page. In contrast to #125 and #127, I did try to introduce a few design changes, since the functionality of this page is simpler and more limited than the tools and chapters pages.

*Note: this is for functionality and layout review, not for code review, so please don't merge yet.*

## Notable changes from view implementation
- **Entry**: Manage Tracks is hidden when there is only one audio file
- **Include/exclude**: Toggling exclude keeps the row in place - I find it less distracting than moving to the top
- **Save**: Stays on the page with a toast and refetch
- **Reset & change detection**: Reset with confirm when there are edits; Save only appears when `hasChanges`
- **Layout**: 
  - Book title header
  - fixed table with scrolling rows
  - Save/Reset at the bottom
  - Below `lg`, compact layout with a **Sort by** dropdown (replacing the sortable columns)
  - Columns with no contents are not displayed
  - Responsive column display (narrower displays show less columns)
- **Sort & drag**: 
  - **Custom order** after manual drag on mobile
  - `Current` sort and mobile `#` column use saved indices without excluded files jumping to the top; 
  - Track-from-metadata sorts disc then track when applicable.
  - Using drag handle icons like other sortable lists.
- **Touch**: 
  - Truncated filenames use tap-to-toggle tooltips; 
  - Larger touch targets on rows, toggles, and drag handles.

Please review the functions and layout (including narrower screens and mobile), and let me know what you think.


advplyr:
I like all of these changes. Big improvement.

One minor thing, I think we should make the disc columns sortable also. I'm pretty sure this was requested for the Vue app. The first audiobook that I opened had all tracks set to 1 and the disc column is incrementing. This can happen with old audiobooks where each audio file was on a different disc and the discs don't have tracks. After disabling a track I wasn't able to get the proper sort without manual drag, but I could have gotten it if the disc columns were sortable.

Besides that I don't have anything else. Is there more you want to polish on this?

advplyr:
I added a cover image thumbnail next to the book title then updated both chapters and tools page to use this. This aligns the headers also.
https://github.com/audiobookshelf/audiobookshelf-client-react/commit/4135fb08a7c8732043ac0e3dc1d09ffb844cd11b
image

Not sure if you touched anything there

mikiher:
> One minor thing, I think we should make the disc columns sortable also. 

Done.

> Besides that I don't have anything else. Is there more you want to polish on this?

No, you can merge from my perspective.

mikiher:
> I added a cover image thumbnail next to the book title then updated both chapters and tools page to use this. This aligns the headers also. 

I didn't touch, just needed to rebase on latest version of master that contained this commit.
This is now fixed. PTAL again.

advplyr:
Great update, thanks!

PR #130, opened 2026-07-04
Podcast Latest page

mikiher:
This migrates the Latest Episodes page of podcast libraries.

### Notable changes from Vue implementation

- **Route**: `/library/{id}/latest` instead of `/library/{id}/podcast/latest`.
- **Layout**:
  - Group cover (up to 4 unique podcast covers in recency order), title, and episode list side-by-side on desktop
    - This makes it similar (but not identical to the episode playlist page)
  - Fix title and description clipping on mobile
- **Per-row play queue**: React correctly skips finished episodes when building the queue
- **Play/pause**: React toggles whenever the episode is streaming (playing or paused
- **Play button labels**: React correctly uses i18n strings
- **Play button label after Mark as Not Finished**: React treats a progress row with `currentTime === 0` as unstarted (plain duration, e.g. `31:49`)
- **Progress bar when finished**: React hides the bar when `isFinished`
- **Descriptions**: React sanitizes HTML so as not make links clickable, and hides empty descriptions
- **Data loading**: React server-fetches on page load
- **Keyboard**: React episode title is a focusable button (Enter/Space opens view modal)

Additionally:
- Refactored some functions and components so that there's more code reuse between `RecentEpisodeRow`, `EpisodeRow`, and `CompliationItemListRow`
- Fixed a couple of bugs in `LazyTooltip`


advplyr:
This is working great, thanks!
I'm holding off on going over the queue stuff until we can see it in the audio player, so I skimmed over that.

PR #133, opened 2026-07-07
Batch actions implemented and wired

mikiher:
This implements and wires the batch edit page and the rest of the other actions.

### Notable changes from Vue

- **Route**: `/library/{id}/batch` (library-scoped)
- **Selection handoff**: Writes a `BatchEditSession` to `sessionStorage` before navigation because `LibraryLayoutWrapper` clears live selection on every pathname change
- **Return after save**: Refirects to the page the user came from (`returnPath` in session, e.g. series detail, author page, `/items`, `/latest`)
- **Episode batch editing**: Added
- **Episode save strategy**: Saves changed episodes **sequentially** via `updatePodcastEpisodeAction` (one request at a time). 
  - Server batch episode endpoint will be added later.
- **Save bar visibility**: Shows the footer only when there are unsaved changes
- **Batch page reset**: Adds a **Reset** button in the save bar with a confirm dialog
- **Apply to all label (React en-us)**: 
  - Changed **Map details** header **Apply to all** 
  - Related strings: **Fill from items**, **Fill from this item**.
- **Apply to all default open**: 
  - Expands the panel on load
- **Mobile layout**: Fixed (was unusable before).
- **Per-item cards**: Stacked column with `gap-4` between bordered cards
- **Border contrast**: React uses `border-foreground/15` on map panel, item cards, and save bar
- **Add to Playlist**: Added for Book libraries (had only Add to Collection until now)

*Note: This PR adds episode batch actions, but these are not supported in the server API as well as library item batch actions. In this PR, we use existing sequentially-called single-episode API. When the server supports batch api for episodes, we should switch to those*


mikiher:
Also implemented the following:
- Big refactor of internal-api routes that shortened most `route.ts` files to bare minimum.
- All the rest of the batch actions are now wired and working.
- Fixed text selection on shift-click in selection mode

advplyr:
I still have some more to test on the batch actions but I want to merge this because the settings page is broken on master right now without this PR.
Thanks!

PR #134, opened 2026-07-09
Keep library home shelves in sync with socket events

mikiher:
This keeps personalized home shelves live-updated via socket events and fixes empty-home UX when shelves are empty but the library still has items.

### Notable changes
- Socket listeners update personalized shelves on item/author/episode add-remove events
- Hide empty shelf rows; show scan empty state only when the library has no items
- Show browse-to-items message when the library has items but no home shelves

*Note: This depends on [server PR #5354](https://github.com/advplyr/audiobookshelf/pull/5354)*

### DIFF
--- a/src/app/(main)/library/[library]/(book)/authors/[author]/AuthorClient.tsx
+++ b/src/app/(main)/library/[library]/(book)/authors/[author]/AuthorClient.tsx
@@ -13,7 +13,7 @@ import { useSocketEvent } from '@/contexts/SocketContext'
 import { useUser } from '@/contexts/UserContext'
 import { useTypeSafeTranslations } from '@/hooks/useTypeSafeTranslations'
 import { filterEncode } from '@/lib/filterUtils'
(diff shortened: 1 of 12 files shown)


advplyr:
I think the no-home-shelves is unnecessary and we should always be populating the home page. I'm not sure it is possible to have an empty home page but have library items.

I have been meaning to update the recently added shelves to always pull the most recent 10 items.

I'll figure that out later, thanks!

mikiher:
> I think the no-home-shelves is unnecessary and we should always be populating the home page. I'm not sure it is possible to have an empty home page but have library items.

It is definitely possible on podcast libraries currently. If the library hasn't been active for a long time (no recent episodes or podcasts), then the homepage becomes empty. I have such a library.

> I have been meaning to update the recently added shelves to always pull the most recent 10 items.

Sure, that will resolve this. I think this requires a server change, though.

> I'll figure that out later, thanks!

PR #137, opened 2026-07-11
Single series page

mikiher:
This migrates the single-series page to use BookshelfClient (Vue parity) and wires the remaining toolbar actions.
- Series page
  - Replaces flat SSR book grid with paginated BookshelfClient
  - Overrode standard bookshelf items query with series filter query
  - Implemented collapse/expand sub-series
  - Implemented mark series finished/unfinished
- BookshelfClient
  - Added queryOverride / registerToolbar props
  - Fixed duplicate items when page size changes
  - Added progress on collapsed sub-series cards

PR #138, opened 2026-07-12
Fix player & player queue gaps

mikiher:
This fixes most of the existing gaps from the Vue implementation.

### Notable changes from Vue
- Player queue
  - queue updated on socket remove/updated events
- Player queue modal
  - Pause and Remove icon button on currently playing item
    - Remove advances to next item when auto-play is on
  - Title links to book page or episode details modal (with nav context)
  - Keyboard tab navigation supported, with tab stops on Title link, Play and Remove
- Player UI
  - Next tooltips for no-chapters updated to "Next Episode in Queue"/"Next Book in Queue"
  - Previous button for no-chapters behavior updated
    - First click goes to start of item
    - Second click within 3 seconds goes to previous item
  - Cover-derived accent backdrop
- Player
  - Player rate persisted to local storage instead of user settings

### The follwoing Vue features were implemented
- Chromecast support
- Podcast metadata details displayed on player bar
- Media session API for lock screen display
- HLS error fallback
- HLS buffer progress on track bar

*Note: The last two HLS features were tested by blocking `*://*/public/session/*/track/*` on DevTools Network Request Conditions*

### Not yet implemented
- Chapters Hotkey (L) - do we really need this? it's totally undiscoverable
- Player mobile layout

*Note: this depends on #137 - please do not merge until I rebase to master*


src/app/(main)/library/[library]/item/[item]/LibraryItemClient.tsx
```diff
@@ -39,6 +39,11 @@ export default function LibraryItemClient({ libraryItem: initialLibraryItem }: L
   const [isEditModalOpen, setIsEditModalOpen] = useState(false)
   const [isCoverEditModalOpen, setIsCoverEditModalOpen] = useState(false)
   const [isClearQueueDialogOpen, setIsClearQueueDialogOpen] = useState(false)
+  const podcastEpisodesInOrderRef = useRef<PodcastEpisode[]>([])
```
Comment (advplyr) [inline src/app/(main)/library/[library]/item/[item]/LibraryItemClient.tsx:42]: Keeping the state here causes the component to re-render anytime the sort/filter changes. To prevent the unnecessary re-render of this component I switched to a ref and getter strategy.

Comment (advplyr) [reply src/app/(main)/library/[library]/item/[item]/LibraryItemClient.tsx:42]: This made me look deeper into state causing re-renders and the playerHandler state gets updated every tick while audio is playing. So anything using useMediaContext will re-render a few times a second.

This stems from something I set up when building the media player without understanding react well enough. I'm looking into a fix for this but I'm sure we'll find a few other places with the same problem.

Comment (advplyr) [reply src/app/(main)/library/[library]/item/[item]/LibraryItemClient.tsx:42]: I cleaned up the player here https://github.com/audiobookshelf/audiobookshelf-client-react/pull/138/changes/58ccfa362bb8e089d591fe7bc08dfa6a269a24ae

usePlayerState is separate now

advplyr:
I still have to test chromecast and HLS. I'll do that when I'm back at my office this weekend or early next week.

For the hotkeys I think we should be putting the hotkey inside the tooltip. "Chapters (L)" or something more clear. We could add a hotkeys modal on the account page later.

Thanks!

PR #142, opened 2026-07-15
Media card fixes

### LINKED ISSUE #105: Context menu remains open when clicking the media card edit button
1. Click a media card context menu
2. Click that media card's edit icon button

Expected: context menu should be closed and edit modal should open
Actual: context menu remains open and edit modal opens.

### LINKED ISSUE #141: Editing a book on the home page causes the series number to show
If you book on the home page has a series, then when making any edit to the book the series number will appear on the media card after save.

Image

mikiher:
This fixes #105 (again).

The isAdditionalInside predicate was incorrect and included overlay actions on every card. Now it includes only the card itself as it should.

This also fixes #141
- Card no longer infers sequence from metadata.series[]
- For the continue-series shelf cards, item updates preserve server-provided metadata.series (which has a different shape)

### DIFF
--- a/src/components/widgets/media-card/BookMediaCard.tsx
+++ b/src/components/widgets/media-card/BookMediaCard.tsx
@@ -13,11 +13,13 @@ export default function BookMediaCard(props: BookMediaCardProps) {
   const { libraryItem } = props
   const media = libraryItem.media as BookMedia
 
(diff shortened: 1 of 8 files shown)


advplyr:
nice, thanks!

PR #145, opened 2026-07-18
Schedule automatic podcast downloads

mikiher:
This migrates what used to be the **Schedule** tab for podcasts. In React this is added as a separate modal, wired as a podcast context menu item.

### Notable changes from Vue
- Enable flow: Vue uses an **Enable** checkbox that gates form fields; React treats **Save** as enable and adds a separate **Disable automatic downloads** button when already enabled
- Save payload: always sends `autoDownloadEpisodes: true`, `autoDownloadSchedule`, and both limit fields via `updateLibraryItemMediaAction`
- Disable payload: `{ autoDownloadEpisodes: false }` only
- Default cron: `0 0 * * *` when not enabled or no saved schedule
- Cron UI: `CronExpressionBuilder` + separate `CronExpressionPreview` (Vue embeds preview in the builder)
- No RSS feed: 
  - show warning alert when `feedUrl` doesn't exist
  - disable on footer when enabled but feed is missing

### DIFF
--- a/src/app/(main)/components_catalog/examples/CronExpressionBuilderExamples.tsx
+++ b/src/app/(main)/components_catalog/examples/CronExpressionBuilderExamples.tsx
@@ -117,8 +117,6 @@ export function CronExpressionBuilderExamples() {
               </Code>{' '}
               - the server language, date, and time settings
             </span>
-            <Code className="px-1">className?</Code>
-            <span>classes to merge with the default classes</span>
           </div>
         </div>
       </ComponentInfo>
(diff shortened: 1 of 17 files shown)


advplyr:
With this design it is unclear that it is enabled. If I open the schedule modal and see that it has an interval filled out then I would assume it is configured and close the modal.

The design in Vue is not great but it is clear that it is not enabled.

advplyr:
I think changing the button to say "Enable" will suffice for now

image

advplyr:
I realized that we never implemented the change we talked about for getting the server timezone. The server timezone is how we can accurately show the next run datetime.

I decided to update the server settings to include the timezone: https://github.com/advplyr/audiobookshelf/commit/08bbf9cd9d22df00895c94749e1283f63bb861a2

Then I updated the components to use that timezone. I realized there are more bugs there and I don't want to include them in this PR.

- The `cron.ts` file has its own `formatDate` function (why not use datefns?) and it is not properly using the server timezone. It is only appending the timezone string to the date that is using the browser timezone
- The server time format always uses 12 hour time even when 24 hour time is selected (regex bug)

advplyr:
Thanks!

PR #146, opened 2026-07-19
Fix users page gaps

mikiher:
This updates the **Users** settings area to fix remaining gaps from Vue: the users list, user detail page, and per-user listening sessions view. 

### Notable changes from Vue

- **Users list data**: loaded via server `getUsers('include=latestSession')` in `page.tsx` and passes props to `UsersClient` / `UsersTable`
- **User detail listening load**: listening sessions and stats are loaded in **parallel**
  - this saves about 400-500 ms in comparison to sequential.
  - queries seem not to be as efficient as possible on server, noted to investigate this later
- **Listening stats loading UI**: shows `UserListeningStatsSkeleton` until loaded
- **Legacy API token**: Label uses `LabelLegacyApiToken` (was hardcoded English); warning uses `t.rich` + Next `Link` to `/settings/api-keys` instead of `v-html` with a raw `<a>`
- **Back link on detail**: `LabelBackToUsers` 
- **Total time listened**: Single line with optional today inline — e.g. `3h 20m (20m today)`

Also fixed scrollbar visible issue in all settngs pages


advplyr:
Looking good. The goal is to drop the legacy tokens when we release this. Thanks!

Issue #112, opened 2026-05-14
remaining todo list


advplyr:
## major
- [x] playlist page
- [x] collection page
- [x] chapter edit page
- [x] audio tools page (embed/encode)
- [x] audio tracks page
- [x] podcast latest page
- [x] podcast add page
- [x] podcast queue page
- [x] podcast schedule auto downloads modal
- [x] e-reader
- [x] library stats page
- [x] library item edit cover modal
- [x] audio player casting
- [x] audio player sleep timer
- [x] audio player bookmarks
- [x] audio player queue & add to queue for books/episodes
- [x] audio player mobile layout
- [x] podcast episode edit modal
- [x] bulk selecting library items & bulk actions
- [x] bulk library item edit page
- [x] search page (when pressing enter on search)
- [x] settings: email
- [x] settings: notifications
- [x] settings: authentication
- [x] migrate/set up i18n (#156)
- [x] release notes modal (when clicking version in side rail)
- [x] OIDC login (#155)
- [x] redesign action buttons on book page tables
- [x] PWA
- [ ] figure out how to support custom base path (in progress - issue #217)

## minor
- [x] Library dropdown in the app bar (#150)
- [x] socket event listeners on home page to auto update shelves
- [x] listening sessions page start playback from timestamp
- [x] podcast library export OPML (in library page context menu)
- [x] series page collapse sub-series
- [x] series page re-add to continue series
- [x] settings page purge cache buttons (maybe do this better)
- [x] home page remove from continue listening & remove from continue series
- [x] wire play button on item page (#153)
- [x] add progress info on item page  (#153)
- [x] fix mobile seachbox z-order  (#150)
- [x] fix button text wrapping on mobile account page  (#153)
- [x] add item count on playlists, collections, narrators toolbar  (#153)
- [x] remove toolbar from library stats  (#153)
- [x] fix redirect url after creating playlist from a collection  (#153)
- [x] wire edit icon on playlists and collections pages  (#153)
- [ ] look into collection/playlist page cover preview styling
- [x] podcast page doesn't update after rescan ([server pr #5409](https://github.com/advplyr/audiobookshelf/pull/5409))
- [x] When disabling chromecast support the chromecast icon only disappears after reload  (#153)
- [x] Simple text input metadata fields are not trimmed before saving (on book edit)  (#154)
- [x] in series page, if the current filter has no series, the filter is removed from the toolbar  (#153)
[...]

PR #150, opened 2026-07-22
Mobile library side rail & AppBar polish

### LINKED ISSUE #110: Implement side bar on mobile
Mobile doesn't currently have a side bar (or equivalent).

I don't think we have yet agreed on a design (though I *think* we agreed the tab interface is not a good choice).

mikiher:
This adds a mobile slide-in drawer for library navigation, toggled from the **AppBar logo** when a library is selected. In Vue the side rail is desktop-only (`hidden md:block`) with no mobile alternative.

This fixes #110 

### Notable changes from Vue

- **Mobile navigation**: New **`SideRailMobileDrawer`** with backdrop dismiss, library header (icon + name), and horizontal nav links
- **Entry point**: AppBar logo toggles the drawer on `< md`; desktop logo remains a home link
- **Shared nav**: Extracted **`SideRailContent`** from **`SideRail`** for desktop rail and mobile drawer (`variant: 'rail' | 'drawer'`)
- **Library scope**: Uses the effective library from the dropdown, so nav is available wherever Appbar is visible.
- **Close behavior**: Drawer closes on backdrop tap, back arrow, nav link, route change, and resize to desktop

### Additional AppBar polish fixes:
- **Library icons**: Show each library’s icon in the dropdown button and menu
- **Hide selected library in menu**: Keep the active library in the button, but omit it from the open menu (Vue parity).
- **Portaled menu alignment**: Anchor the menu to the full control wrapper so the portaled list lines up with the selected value.
- **Focus outline**: Stop clipping the libraries dropdown focus ring in the AppBar.
- **AppBar logo buttons**: Use `ButtonBase` for the mobile drawer toggle and desktop home link so focus styles render correctly.
- **Stable desktop width**: Size the dropdown to the longest library name (capped at `max-w-52`) so menu items aren’t truncated when a shorter library is selected.
- Fixed bug where the libraries dropdown disappeared when the selected library was deleted.
- Added empty state (no libraries) on /settings/libraries, to match Vue
- Refactored and fixed AppBarNav keyboard navigation and click outside behavior.
- **Global search** on mobile:
  - Render search results menu in a portal so it is not clipped by the app bar’s `overflow-x-hidden`
  - When the search icon expands the field, use `flushSync` and focus the input via ref other wise the focus would be lost

### DIFF
--- a/cypress/tests/components/app/AppBarNav.cy.tsx
+++ b/cypress/tests/components/app/AppBarNav.cy.tsx
@@ -1,0 +1,85 @@
+import AppBarNav from '@/app/(main)/AppBarNav'
+import { AppRouterContext } from 'next/dist/shared/lib/app-router-context.shared-runtime'
+import type { AppRouterInstance } from 'next/dist/shared/lib/app-router-context.shared-runtime'
+import * as navigation from 'next/navigation'
+import { ReactNode } from 'react'
+
+function mountAppBarNav(props: { userCanUpload?: boolean; isAdmin?: boolean; username?: string } = {}, options?: { alignEnd?: boolean }) {
+  const router = {
+    back: cy.stub(),
+    forward: cy.stub(),
+    refresh: cy.stub(),
+    push: cy.stub(),
+    replace: cy.stub(),
+    prefetch: cy.stub()
+  } as AppRouterInstance
+
+  cy.stub(navigation, 'useRouter').callsFake(() => router)
+
+  const appBarNav = <AppBarNav username={props.username ?? 'testuser'} isAdmin={props.isAdmin ?? false} userCanUpload={props.userCanUpload ?? false} />
+
+  const ui: ReactNode = (
+    <AppRouterContext.Provider value={router}>{options?.alignEnd ? <div className="absolute end-0">{appBarNav}</div> : appBarNav}</AppRouterContext.Provider>
+  )
+
+  cy.mount(ui)
+
+  return router
+}
+
+describe('<AppBarNav /> desktop keyboard navigation', () => {
+  beforeEach(() => {
+    cy.viewport(1024, 768)
+  })
+
+  it('opens the menu with Enter on the desktop username button', () => {
+    mountAppBarNav()
+    cy.get('button[aria-haspopup="menu"]').focus()
+    cy.get('button[aria-haspopup="menu"]').type('{enter}')
+    cy.get('[role="menu"]').should('be.visible')
+  })
+
+  it('moves focus between menu items with arrow keys', () => {
+    mountAppBarNav()
+    cy.get('button[aria-haspopup="menu"]').focus().type('{enter}')
+    cy.get('[role="menuitem"]').eq(0).should('have.focus')
+    cy.focused().type('{downarrow}')
+    cy.get('[role="menuitem"]').eq(1).should('have.focus')
+    cy.focused().type('{uparrow}')
+    cy.get('[role="menuitem"]').eq(0).should('have.focus')
+  })
+
+  it('closes the menu with Escape and returns focus to the trigger', () => {
+    mountAppBarNav()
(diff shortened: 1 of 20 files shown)


modarah:
Hi @mikiher!

Out of curiosity I checked out your branch and I noticed that the backdrop in `SideRailMobileDrawer` and `SideNavMobileDrawer` has `z-40` but the `AppBar` has `z-60` and `AppBarSelectionOverlay` has `z-70` which keeps the app bar tappable/clickable and could cause this:
grafik

Navigating with keyboard is better in the menu though I noticed if the menu is open and you press tab (without closing it) the focus will go back to the AppBar logo and link

Another nitpick:
Entry point for mobile navigation through the AppBar logo is very unexpected. But I don't have a better suggestion :D

mikiher:
> Hi @mikiher!
> 
> Out of curiosity I checked out your branch and I noticed that the backdrop in `SideRailMobileDrawer` and `SideNavMobileDrawer` has `z-40` but the `AppBar` has `z-60` and `AppBarSelectionOverlay` has `z-70` which keeps the app bar tappable/clickable and could cause this: 

Thanks, fixed.
 
> Navigating with keyboard is better in the menu though I noticed if the menu is open and you press tab (without closing it) the focus will go back to the AppBar logo and link

Thanks, fixed. I changed the AppBarNav focus to stay on the trigger like other menu controls, so now tab just does what it should do.

> Another nitpick: Entry point for mobile navigation through the AppBar logo is very unexpected. But I don't have a better suggestion :D

As I mentioned in #112, the mobile appbar design is still in flux. This is my take on resolving the mobile library navigation (we wanted to move away from the Vue horizontal tab bar, which is cluttered and waists vertical space). I think the drawer solution is good even if somewhat unusual (plus the logo is known as a navigational link to existing users), and putting it under the audiobookshelf button makes the drawer accessible from everywhere in the app. Let's see what @advplyr thinks of it - I'm open to changing this.

modarah:
Thank you for fixing!

As you said, the logo is a navigation link so even if it's not expected, users will evantually tap it and learn (like I did 😁).

My take on it isn't better tbh because of different reasons but mainly... moving the logo 😅


Screenshot 2026-07-25 at 22 03 47
<br>
or
<br>
Screenshot 2026-07-25 at 22 00 55

mikiher:
> My take on it isn't better tbh because of different reasons but mainly... moving the logo 😅

Yeah, I don't think two separate icons are such a good idea... the appbar is already too cluttered as it.

I have some ideas on a design that resolves all the current issues, but I need to play with it a bit. I'd think of the current design more as an imperfect V1.

modarah:
> Yeah, I don't think two separate icons are such a good idea... the appbar is already too cluttered as it.
>
Yeah, that is one of the other reasons I assumed. Though, personally I don't find it cluttered in comparison to the vue one. Here, I know exactly what to expect, menu, search, libraries dropdown, app logo (home), side bar trigger, but as I said that's a personal taste.
>
> I have some ideas on a design that resolves all the current issues, but I need to play with it a bit. I'd think of the current design more as an imperfect V1.
>
Exactly!

As always, thank you for the work!

advplyr:
> Entry point for mobile navigation through the AppBar logo is very unexpected. But I don't have a better suggestion :D

I agree. I think we should remove the app icon on mobile and replace it with a menu icon.

I want to get this merged so I can continue going through the stack of PRs, but we can continue to discuss this.

This is what I went with:

image

One negative with this design is the settings page gets a double menu. 

image

---

Separate but kind of related note, I left the library dropdown off of the app bar on the settings pages intentionally. I thought the Vue implementation was not good design because when you are in the settings pages you are no longer library specific. Settings pages should have nothing to do with what the selected library is.

I think the Vue app did this poorly of trying to show a selected library, while at the same time having some pages that affect all libraries (like in item metadata utils).

tldr; the settings section of the app shouldn't be showing a selected library even though it can be convenient to jump libraries from there, it is not good design.

PR #151, opened 2026-07-23
Differentiate podcast fetch vs device download

### LINKED ISSUE #147: Download function in podcast context menu unclear
I realize that the Download podcast context menu item is unclear and potentially misleading.

What it does currently is immediately start to download all the "downloaded" (note the double meaning here) episodes as a zip file to the client, without any clear indication (except on the browser's Downloads button on the top right, which isn't that visible).

1. It's quite confusing because of the dual meaning of "download" (i.e. download to server vs. download to client)
2. It also starts a potentially very large download without any warning.

I think we should either change the menu item label and/or append a confirmation dialog, which should also display the (estimated?) download size.

mikiher:
Fixes #147

Clarifies the dual meaning of “download” for podcasts - server-side episode fetching uses **fetch** wording (this seems clearer than "get" suggested in the Bug), while zip/file downloads to the user’s device stay **download**.

### Changes

- **Renaming**: Renamed server-side UI strings in `en-us.json` (fetch queue, auto-fetch, etc.); client downloads keep “download”
- **Confirmation**: Podcast show zip downloads now prompt before starting, with estimated size when available (media card menu and batch bar)
- **Helper**: Shared `openPodcastDeviceDownloadConfirm` in `podcastDownload.ts`

*Note: this depends on #150. Please do not merge before I rebase to master.*


advplyr:
The term "fetch" works great for me, but that may be because I'm familiar with the javascript fetch API. We'll find out during the testing phase I'm sure.

The confirm prompt is a nice improvement. Thanks!

PR #152, opened 2026-07-29
Fix local docker build

modarah:
While building the local docker image I got this error `failed to solve: process "/bin/sh -c pnpm install --frozen-lockfile" did not complete successfully: exit code: 1`


mikiher:
Can you make those lines identical to the ones in `Dockerfile`, so as not to deviate from the source of truth too much?
```
COPY ./client-react/scripts/sync-pdfjs-vendor.mjs ./scripts/sync-pdfjs-vendor.mjs
COPY ./client-react/scripts/sync-unrar-wasm.mjs ./scripts/sync-unrar-wasm.mjs
```

The longer term solution, though, would be to delete `Dockerfile.local`, use only `Dockerfile`, and use build contexts to pass the abs-server directory (in both `docker-compose.yml` and `docker-build.yml`)

modarah:
It will not work because the context in `Dockerfile.local` is different, we are already in `client-react` . In [docker-build.yml](https://github.com/audiobookshelf/audiobookshelf-client-react/blob/master/.github/workflows/docker-build.yml#L23-L32) the CI has a different tree before building.

If you want, I can adapt the build contexts to get rid of `Dockerfile.local` like you mentioned. We already have one for the server, then I can add `abs-client`.

mikiher:
Yeah, sorry, identical will not work - just similar.

If you want to do the full removal of `Dockerfile.local`, please go ahead, I think it's a better solution. Just make sure you test both `docker-compose.yml` and `docker-build.yml`.

modarah:
I tested `docker-compose.yml` successfully and ran `buildx` with the build contexts too but I'm not sure how can I run `docker-build.yml`

mikiher:
You can run it in your Github fork.  
1. Go the Actions tab. 
2. Select "Build and Push Docker Image"
3. Click "Run Workflow" on the right
4. Change the Docker tag to some test name
5. Click the green "Run Workflow" button.

You might not have permissions to login to ghcr so you might need to comment out some steps for your testing and then re-add them. What's important is to verify that the build completes successfully - you don't need to test the push.

modarah:
Thank you for the instructions!

Take a look at this [job's log](https://github.com/modarah/audiobookshelf-client-react/actions/runs/30655953332/job/91240298277). Buidling was successful, job failed because of pushing the image (I wanted to see what will happen :D) 

Edit:
Here is a [successful](https://github.com/modarah/audiobookshelf-client-react/actions/runs/30660066220/job/91253917863) one after I set push to false

mikiher:
Very nice, thanks!

PR #153, opened 2026-07-30
Various fixes

mikiher:
This contains various fixes to issues in #112.

### Fixes

- Wire library item page cover play button
  - Move play logic to `useLibraryItemPagePlay`
  - On podcast page, Play shows `MessageNoEpisodesToPlay` toast if no episodes match filter or search (Vue does nothing, which can be confusing)
- Add library item progress info panel (Vue parity)
  - Item header order: details → podcast download banners → progress panel → action buttons
  - Panel is start-aligned on mobile (Vue centers with `mx-auto`)
  - Panel and reset button accessibility:
    - Reset button is tabbable
    - Reset button has larger hit/touch target (but looks the same)
    - Panel has `role="region"` and aria-title
    - Reset button has `ButtonResetProgress` localized aria-label (Vue had fixed English string)
- Fix account page actions buttons wrapping on mobile.
- Show toolbar item count on mobile as a compact badge on all library pages
  - Include narrators page (count synced from `NarratorsClient`)
  - Filter/sort grow equally into remaining toolbar space on small screens
  - Wait for item count before showing filter/sort so mobile widths don’t jitter
  - Dropdown `wrapText` option so menu items wrap (used by library filter/sort)
  - Submenus use Floating UI flip/size to stay fully on-screen (so the scrollbar isn’t clipped)
- Hide toolbar on library stats page
- Fix redirect URL after creating a playlist from a collection on the collection page
- Wire collection/playlist card edit buttons
- Fix hydration mismatch in `useVersionData`
- Keep server settings in sync app-wide after admin changes (similar to vue $store)
  - `UserContext` exposes `mergeServerSettings` to replace `serverSettings` on the login payload
  - Main settings (`SettingsClient`):
    - read `serverSettings` from context (no local state)
    - optimistic `mergeServerSettings` on save/language change with revert on error
    - merge API `serverSettings` after successful PATCH
  - Similar changes in Backups and logs settings
  - Settings pages no longer pass `serverSettings` into clients that only need live context data
  - `settings/actions.ts` uses a file-level `'use server'` so client components can import server actions safely
- Chromecast launcher reacts when Chromecast support is toggled off in settings (no full page reload)
  - `ChromecastLauncher` gates on `serverSettings.chromecastEnabled` from context
  - `ChromecastContext` clears init/casting state and notifies cast session inactive when support is disabled
- Keep series page filter/sort in toolbar when a filter returns no results
  - `Toolbar` uses `seriesFilterBy` (not `filterBy`) to detect an empty series bookshelf
- Fix uncaught error when removing the last item from a playlist on the playlist page
  - Backend deletes empty playlists; client now navigates to the playlists list instead of refreshing the detail page
  - `PlaylistClient` listens for `playlist_removed` and redirects away (Vue parity)

*Note: this depends on #151. Please do not merge before I rebase to master.*

### DIFF
--- a/cypress/tests/components/ui/DropdownMenu.cy.tsx
+++ b/cypress/tests/components/ui/DropdownMenu.cy.tsx
@@ -328,5 +328,22 @@ describe('<DropdownMenu />', () => {
       cy.get('[role="listbox"] > li').eq(1).should('contain.text', 'Short')
       cy.get('[role="listbox"] > li').eq(1).should('contain.text', 'This is a very long text')
     })
+
+    it('wraps long text when wrapText is set', () => {
+      const longText = 'This is a very long text that should wrap onto multiple lines in the dropdown menu item instead of truncating with an ellipsis'
+      const longItems: DropdownMenuItem[] = [{ text: longText, value: 'long1' }]
+      cy.mount(
+        <div className="w-48">
+          <DropdownMenu {...defaultProps} items={longItems} wrapText />
+        </div>
+      )
+      cy.get('[role="listbox"] > li')
+        .eq(0)
+        .find('span')
+        .first()
+        .should('have.class', 'line-clamp-2')
+        .and('have.class', 'break-words')
+        .and('not.have.class', 'truncate')
+    })
   })
 })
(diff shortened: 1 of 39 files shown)


advplyr:
Thanks!

PR #154, opened 2026-07-30
Fix text field trimming behavior across the code

mikiher:
## Summary

Trailing/leading whitespace in text fields could reach save actions or leave Save enabled with no real change, because  `Btn` uses `mousedown` `preventDefault` (which is the desired behavior and matches Vue), so focused inputs do not blur on click and blur-only cleanup never runs.

This PR applies a consistent fix across forms:

- **Handler-level trim** — normalize strings when building save payloads and when comparing dirty state / validation
- **Blur cleanup (Vue parity)** — opt-in `trimWhitespace` on `TextInput` / `TextareaInput`
- **`useDetailsEdit`** — new `trimFields` option for book/podcast item editors

### Areas updated

- **Shared:** `TextInput`, `TextareaInput`, `useDetailsEdit`
- **Item editors:** book, podcast, episode details
- **Modals:** author edit, new podcast, RSS feed open
- **Settings:** library edit, API key create, notification templates, user account
- **Widgets:** batch map details, metadata match apply, cover URL/search

*Note: this depends on #153. Please do not merge before I rebase to master.*

### DIFF
--- a/src/app/(blank)/login/ServerInitForm.tsx
+++ b/src/app/(blank)/login/ServerInitForm.tsx
@@ -38,7 +38,7 @@ export default function ServerInitForm() {
       try {
         const payload = {
           newRoot: {
-            username,
+            username: username.trim(),
             password
           }
         }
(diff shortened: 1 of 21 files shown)


advplyr:
Thanks!

PR #158, opened 2026-08-03
Issues page

mikiher:
Migrate the Issues page and side rail link.

### Notable changes from Vue

- **Route**: uses a dedicated `/issues` page (`libraryIssuesPage.ts`). 
- **Filter dropdown on issues page**: hides `LibraryFilterSelect` on `/issues` (`ItemsToolbarExtras`)
- **Remove-all confirm**: uses `ConfirmDialog` with i18n strings
- **Remove-all button label**: uses ICU plurals `ButtonRemoveIssuesBooks` / `ButtonRemoveIssuesPodcasts` with `{count}`. 
- **`numIssues` source**: reads `numIssues` from `GET /api/libraries/{id}/filterdata` 
- **Live `numIssues` updates**: 
   - refetches `/filterdata` after scans, rescans, deletes, and remove-all; 
   - reconciles from the issues bookshelf on `/issues`
- **Empty issues redirect**: uses `useLibraryRouteGuard` redirects to library home and resets `filterBy` to `all` when `numIssues === 0` on `/issues` 
- **Downloads for broken items**: hides Download on item pages / cards / episode rows when `isMissing` or `isInvalid`.


advplyr:
In the Vue implementation I didn't like that the issues page was a separate page but also a filter on the items page. It also made this messier to migrate to React so I simplified it by removing issues from the filter dropdown. It shouldn't be necessary there when you have a red button in the side rail. Plus we weren't showing the toolbar remove items with issues button when filtering by issues on the item page.

By removing issues from `filterBy` the issues page now doesn't set the filter at all. Instead the `useBookshelfQuery` checks if it is the issues page and overrides the filter when building the API request.

I also updated the issues page to show for admins only.

Thanks!

PR #165, opened 2026-08-04
Fix clipped focus ring on Narrators table remove buttons

### LINKED ISSUE #159: Focus ring on Narrators table remove buttons is clipped
**Repro**
1. Go to Narrators page
2. Navigate to one of the remove buttons using the keyabord

**Expected**: the focus ring should be fully visible.
**Actual:**
Image

Div-47:
## Summary

Fixes the focus ring being clipped on the remove buttons in the Narrators table.

## Changes

- Adjusted the styles to prevent the focus ring from being clipped.
- Preserves the existing layout and behavior.

## Testing

- Verified keyboard navigation.
- Confirmed the focus ring is fully visible.

Fixes #159

### DIFF
--- a/src/components/ui/EditList.tsx
+++ b/src/components/ui/EditList.tsx
@@ -183,7 +183,7 @@ export default function EditList({ items, onItemEditSaveClick, onItemDeleteClick
                     </td>
                   )}
                   <td className="w-1/4">
-                    <div className="flex justify-end pr-2">
+                    <div className="flex justify-end pe-2">
                       <IconBtn
                         size="small"
                         borderless={true}
--- a/src/components/ui/EditList.tsx
+++ b/src/components/ui/EditList.tsx
@@ -183,7 +183,7 @@ export default function EditList({ items, onItemEditSaveClick, onItemDeleteClick
                     </td>
                   )}
                   <td className="w-1/4">
-                    <div className="flex justify-end">
+                    <div className="flex justify-end pr-2">
                       <IconBtn
                         size="small"
                         borderless={true}


mikiher:
1. We prefer using `ps-*` and `pe-*` tailwind classes instead of `pl-*` and `pr-*`, so that we can eventually fully support RTL languages.
2. Since you changed `EditList`, did you test the other components using it?

mikiher:
Also, when you mark a PR as "Ready for Review", please also assign me as a reviewer so I get notified.

Div-47:
ok, let me do that.

Div-47:
Thanks for the feedback. I’ve updated the Tailwind classes from pl-*/pr-* to ps-*/pe-* for RTL compatibility.

I tested the other components using EditList, including the available flows for Custom Metadata Providers, API KEYS Narrators, Backups under settings etc. Everything I was able to test is working as expected. I wasn’t able to test Tags and Genre because I couldn’t add new entries there, but the existing functionality and other components are working correctly.

mikiher:
Genres/Tags get added automatically when you add them to books/podcasts via the edit button

Div-47:
Screenshot 2026-08-05 at 7 13 44 PM
Screenshot 2026-08-05 at 7 13 34 PM
Yeah, its working.

Div-47:
@mikiher I dont have any options to assign a reviewer nor have permission to change the status to Ready for Review. I think i need an contributer tag so that i will have the permissions to do the same.

mikiher:
Thanks!

PR #166, opened 2026-08-04
Server init form

mikiher:
Migrate the first-run server setup flow (`/login` when the server has no root user).

### Notable changes from Vue

- **Language on init**: adds `LanguageDropdown` on the init form; 
  - seeds the `language` cookie from `Accept-Language` via `matchAcceptLanguage` before any cookie exists
- **Language persistence**: persists the init language choice on first admin/root login with `PATCH /api/settings` 
- **Init i18n**: fully translates init copy (title, section headers, labels, confirms, errors) instead of Vue's hardcoded English strings.
- **No-password confirm**: uses `ConfirmDialog` + `MessageConfirmCreateRootUserNoPassword` instead of native `confirm()`.
- **Stale session after DB wipe**: proxy clears `access_token` / `refresh_token` on `/login` when the server is uninitialized
  -  `redirectToLogin` clears both cookies so valid-looking JWTs do not cause an endless loop between `/login` and `/library`.
- **Post-init navigation**: uses `window.location.replace('/login')` after init so stale `?error=` query params from pre-init refresh failures are dropped
  - root with no libraries lands on `/settings/libraries` (`getUserDefaultUrlPath`), similar to Vue behavior


advplyr:
I removed the directory paths. I think they are unnecessary and they caused the server init form to look cluttered especially with the addition of the server default language. The server default language is a nice touch.

Updated verion:

image

Thanks!

PR #167, opened 2026-08-05
Do not show libraries menu when no other libraries exist

### LINKED ISSUE #164: Libraries dropdown shows an empty menu when there's only one library
*Note (this bug happens post #150, which hides the selected library in the dropdown menu [8151cd7](https://github.com/audiobookshelf/audiobookshelf-client-react/commit/8151cd752e344bb42c14deb5271a7cefe5ebbb59)).

Repro: 
When the server has just one library:
1. Click on the libraries dropdown

Actual: An empty dropdown menu is shown
Image
Expected: Don't show menu, or show a "No Other Libraries" menu item

boqiaok:
## Summary

Do not open the libraries menu when no other libraries are available.

## Problem

[After hideSelectedInMenu was introduced, a server with only one library produced an empty menu:](https://github.com/audiobookshelf/audiobookshelf-client-react/issues/164)

When there is only one library, clicking the library selector opens an empty menu.

## Solution

Update toggleMenu to return when menuItemsToShow is empty.

## Testing

With one library, clicking the library selector does not open a menu.

With multiple libraries, the current library remains hidden and the other libraries are displayed.

Fixes #164


mikiher:
This feels somewhat off to me - making the dropdown button inert does seem like the right thing to do, but the expand symbol, and the fact that the menu button can be tabbed into, might cause a user to scratch their head.

Since hiding the the selected library in the dropdown was a deviation from Vue behavior, let's wait for @advplyr and others to comment about the new design first before we continue with this fix.

advplyr:
can you merge master on this?

boqiaok:
Merged the latest master.

mikiher:
@advplyr my preference regarding this: modify the dropdown button to be inert (which is what @boqiaok already implemented), but also 
- add tab index -1 
- remove the expand icon from the button.

(and of course there's the option to switch back to the Vue behavior, which I don't like)

advplyr:
I think it's confusing to have a button there that doesn't do anything. When removing the chevron icon it just looks like a button. Adding the prop `hideSelectedInMenu` to the dropdown component, then adding more logic to handle hiding the chevron icon is too much for this one use-case I think.

mikiher:
OK, I'll just comment that this one use case is likely what most users go through when onboarding, and I'd also guess that many of them stay with just one library. That is why I thought optimizing it made sense.

So just to be clear, do you want to revert the `hideSelectedInMenu` prop?

advplyr:
How many libraries people use is a data piece I'd like to get actually. Maybe I'll poll in Discord, but Discord is going to be heavily skewed towards multi-library power users. Either way I think most people have more than one.

But yeah I think we should revert `hideSelectedInMenu` just to keep things simpler. Remember that we have some new designs that the designer made for us and we opted to hold off on anything major there. So we can come back and improve all this on a future pass.

mikiher:
ok, I reverted it. xlosing this PR and the assiciated issue

PR #171, opened 2026-08-05
Fix #161: activate Series/Collection/Playlist cards on Enter key

### LINKED ISSUE #161: Pressing Enter on focused cards doesn't work on Series/Collections/Playlists
Repro:
1. Go to the series/collections/playlists page
2. Navigate to one of the cards using keyboard tabs
3. Press Enter on the focused card

Expected: page should switch to playlist/collection/series page (same as clicking on the card)
Actual: nothing happens

AryaBait7:
Closes #161

  MediaCardFrame (a div, not a button) only activates on Enter if the
  caller passes it an onKeyDown handler. MediaCard already did this for
  book/podcast cards, but SeriesCard, CollapsedSeriesCard, CollectionCard,
  and PlaylistCard only wired up onClick, never onKeyDown — so keyboard
  users couldn't open those cards.

  Added a handleCardKeyDown to each (mirroring MediaCard's pattern) that
  activates the card on Enter.

  Tested manually: keyboard-only Tab + Enter navigation now works on the
  Series, Collections, and Playlists pages.

### DIFF
--- a/src/components/widgets/media-card/CollapsedSeriesCard.tsx
+++ b/src/components/widgets/media-card/CollapsedSeriesCard.tsx
@@ -78,7 +78,7 @@ export default function CollapsedSeriesCard(props: CollapsedSeriesCardProps) {
 
   const handleCardKeyDown = (event: KeyboardEvent) => {
     if (event.defaultPrevented) return
-    if (event.key === 'Enter') {
+    if (event.key === 'Enter' || event.key === ' ') {
       event.preventDefault()
       event.stopPropagation()
       handleCardClick()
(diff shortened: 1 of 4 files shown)


mikiher:
Good call adding the keyDown handler on CollapsedSeries as well.

Since there's no selection mode on either of these cards, I also added space handling that performs the same action.

Thanks!

AryaBait7:
Thank you @mikiher  for the review and for adding the Space key handling .
I would be happy to help out with other issues if there's anything you could point me toward.

PR #174, opened 2026-08-08
i18n: replace hardcoded strings, fix message anti-patterns, and guard with CI

### LINKED ISSUE #169: Internationalize all user-facing strings
I just added a script that uses all sorts of heuristics to identify and point out hadrcoded English user-facing strings in the code.

Run using:
```sh
pnpm find-hardcoded-strings
```
It currently finds 116 potential issues when running on master (some are false positives, which is expected).

We should reach 0 and run this in CI for every push.

It is possible to mark false positive findings (so that they can be ignored by the script) in the code by adding `// i18n-ignore` comment on the offending line.

mikiher:
Improves i18n across the React client: fix how messages are composed, replace remaining hardcoded UI strings, and add tooling so regressions are caught in CI.

Fixes #169 

### Hardcoded strings

- Replace remaining hardcoded UI text with `t('')` keys in `en-us.json` across the codebase.
- Improve `find-hardcoded-strings.mjs` (skip Material icon names in `IconBtn` children).
- Add GitHub Actions CI (`lint`, `typecheck`, `find-hardcoded-strings`) and local `pnpm check`.
- Strings hardcoded on purpose are ignored by the script using an `// i18n-ignore` comment

### Message anti-patterns

- Remove nested `t()` calls (no translated fragments passed as parameters to another message).
- Replace stitched phrases (template literals, adjacent fragments, `"Label: " + value`) with single parameterized keys (e.g. `LabelFinishedDate`, `LabelDurationWithValue`, `LabelTextWithSubtext`)
   - *note: I know this creates a lot of changes in the strings file, but this IS the recommended practice for i18n, and we should abide by it*
- Use `t.rich()` where markup belongs in the sentence (help text line breaks, settings footer links).
- Widen `TypeSafeTranslations.rich` and tooltip components to accept `ReactNode` for rich help text.
- Allow translated strings in `LoadingIndicator` labels.


advplyr:
Thanks!

PR #175, opened 2026-08-08
Rename Settings page to General and reorganize sections

### LINKED ISSUE #149: Settings within Settings?
When you go to Settings, the first settings page is... Settings. I think this is rather awkward. 

I suggest we do the folliwing: 
1. call that page `General`
2. move settings away from ithe existing `General` section, and remove it
   - `Store Cover with item` and `Store metadata with item` can probably be put in a `Storage` (or similar) section 
   - I think the `Ignore prefixes when sorting` can be moved to `Display`

Thoughts?

mikiher:
Fixes #149 :
- Renames the main settings page/nav item from "Settings" to "General" to avoid Settings-within-Settings
- Moves store cover/metadata toggles into a new Storage section and ignore-prefixes into Display
- Renames `SettingsClient` to `GeneralSettingsClient`


advplyr:
ThankS!

PR #177, opened 2026-08-09
Fix collection group cover backgrounds

### LINKED ISSUE #172: Collection group cover stretches non-square book cover images
When a collection's group cover contains books with a non-square cover image, they are stretched to fill the alloted square, like this:

Image

It should behave like the playlist group cover, which shows the covers in their original aspect ratio with cover background, like this:

Image

blue219:
## Brief summary

Preserves collection cover artwork without cropping when its aspect ratio does not match the rendered group-cover cell.

## Which issue is fixed?

Fixes #172

## In-depth Description

The group-cover components now compare each image's natural aspect ratio with the actual dimensions of its rendered cell. When an image leaves space in that cell, the component renders a blurred background behind an `object-contain` foreground image; covers that fill the cell keep the existing `object-cover` rendering.

The shared hook also cancels stale image-dimension requests, so rapid collection updates cannot apply an old cover's dimensions to a new entry. Square-grid and collection-group covers now use the same sizing logic.

## How have you tested this?

- ESLint on the changed source and component-test files
- Cypress component test: `CollectionGroupCover.cy.tsx` (5 passing)
- Manual Chrome verification of a square cover displayed in a wide single-cover cell
- `git diff --check`

## Screenshots

image

### DIFF
--- a/cypress/tests/components/widgets/media-card/CollectionGroupCover.cy.tsx
+++ b/cypress/tests/components/widgets/media-card/CollectionGroupCover.cy.tsx
@@ -1,0 +1,105 @@
+import CollectionGroupCover from '@/components/widgets/media-card/CollectionGroupCover'
+import { CardSizeProvider } from '@/contexts/CardSizeContext'
+import type { LibraryItem } from '@/types/api'
+
+const bookWithCover = {
+  id: 'book-1',
+  updatedAt: 1,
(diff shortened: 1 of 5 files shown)


mikiher:
Thanks! Works great. I did a bit more refactoring for DRY.

PR #179, opened 2026-08-09
Refactor EditList component into MetadataEditTable

### LINKED ISSUE #168: EditList should be using SimpleDataTable
`EditList` was introduced before `SimpleDataTable` came into being. 
`SimpleDataTable` provides consistent table look & feel across the codebase, and most tables are using it now.

`EditList` should be re-implemented to be built on top of `SimpleDataTable` and also renamed to `MetadataEditTable` (`EditList` is somewhat vague)

modarah:
Closes #168


mikiher:
I fixed a few small issues I found.

Mobile layout is still buggy. I want it behave well on a 360px wide screen:
- numBooks column is missing for narrators
- Save/Cancel text buttons are not suitable for mobile and get clipped. Change to icon buttons.
- padding on mobile narrators page is too large. We should use `px-2 md:px-8`

modarah:
Thanks! I'll take a look later today!

modarah:
- I thought hidding the numBooks column on mobile was intential. Now it's shown.
- Exchanged save and cancel buttons in edit mode with icon buttons. Not sure if you prefer other icons.
- Adapted text input styling to avoid content shifting.

mikiher:
Thanks! nice touch on preventing shifts. I fixed it a bit so it works on desktop as well.

PR #180, opened 2026-08-09
Standardize useMediaQuery usage

### LINKED ISSUE #178: Standardize useMediaQuery usage
`useMediaQuery` is currently used in various places with various query constants.

Usage should be standardized to use a tighter set of constants, which should all be defined in `useMediaQuery.ts`

modarah:
closes #178


mikiher:
Thanks!

Issue #173, opened 2026-08-07
Suggestion: make Upload a library page


mikiher:
The upload button currently lives on the app bar. When you click, you get a page where you select a library and a folder and then upload to it.

This looks a bit weird - why is upload so special to get a special place on the app bar?

I'm suggesting we make it a library page:
- add an Upload link to the library siderail (opens the library Upload page)
- the page would no longer have library selection dropdown - it would act on the selected library.
- all other controls on the page remain unchanged

What do you think?

Vito0912:
I think the difference is that every sidebar item is not connected to permissions, except for Podcasts. 

I do think it is a good idea to remove the "Library" dropdown and base it on the library selected at the top.

I think uploading is still something special. Personally, I find it strange that adding podcasts and the queue have a permanent place on the sidebar 24/7. It feels to me like having an issues tab there even when there are no issues.

Maybe it could put it with the stats in the user selector at the top?

That is just my input. But would also understand it as a separate tab.

mikiher:
> I think the difference is that every sidebar item is not connected to permissions, except for Podcasts. 

That's not a problem. For users who are not allowed to upload, it would just not appear in the siderail (just like button doesn't appear on the app bar today). As you said "Add" for podcasts is also permission-gated (and is similar in spirit to "Upload")

> I do think it is a good idea to remove the "Library" dropdown and base it on the library selected at the top.
> 
> I think uploading is still something special. Personally, I find it strange that adding podcasts and the queue have a permanent place on the sidebar 24/7. It feels to me like having an issues tab there even when there are no issues.

I'm not sure the analogy with issues is accurate.

> Maybe it could put it with the stats in the user selector at the top?

The app bar layout is still in flux, but I think the end goal for the user menu is to only display user-specific menu items (so Settings, for example, is not supposed to be there on mobile)

> That is just my input. But would also understand it as a separate tab.

Vito0912:
> That's not a problem. For users who are not allowed to upload, it would just not appear in the siderail (just like button doesn't appear on the app bar today). As you said "Add" for podcasts is also permission-gated (and is similar in spirit to "Upload"

I also do not mean that it is a problem, nor did I state that anywhere

nichwall:
I agree that changing it to be in the side rail makes more sense. It might make sense to somehow combine it with the "Add podcast" page, but we would need to think about what it should look like.

PR #182, opened 2026-08-10
Settings and app bar - take 2

### LINKED ISSUE #176: Settings pages redesign
I wanted to designate a separate issue for @advplyr 's last comment on #150 . He wrote:

> Separate but kind of related note, I left the library dropdown off of the app bar on the settings pages intentionally. I thought the Vue implementation was not good design because when you are in the settings pages you are no longer library specific. Settings pages should have nothing to do with what the selected library is.
> 
> I think the Vue app did this poorly of trying to show a selected library, while at the same time having some pages that affect all libraries (like in item metadata utils).
> 
> tldr; the settings section of the app shouldn't be showing a selected library even though it can be convenient to jump libraries from there, it is not good design.

I have a followup question: do you also want the search box to be removed from settings pages? it also depends on the selected library.

mikiher:
This fixes #176 (hopefully...)

Settings are server-scoped, not library-scoped. The app bar until now (and Vue implementation) still showed library chrome on settings pages — dropdown, search, hamburger — which implied a “current library” that doesn’t apply there.

This PR redesigns how you enter, navigate, and leave settings, especially on mobile, and cleans up the app bar so library UI only appears where it belongs. This is inspired by how many mobile apps I've seen handle settings and navigation within settings.

### No library chrome on settings pages app bar
On any /settings/* route the app bar no longer shows the library dropdown, search, library hamburger.
- Desktop: Back + “Settings”; the sidenav still handles section switching. Opening settings goes to /settings/general.
- Mobile: Back + section title (e.g. “General”, “Users”)
  - nested pages go back to their parent, not straight to the library
  - leaving settings prefers the page you came from (series, playlists, items, etc.), not always library home
    - If that path isn’t known (refresh / deep link), it falls back to the last/default library
  - section title is removed from the settings block (shown only on desktop)

### Mobile settings: hub instead of a second drawer
Mobile used to stack a settings hamburger/drawer on top of the library side rail — two menus, unclear hierarchy.
Now:
- Settings opens a hub at /settings (card-style links, like Item Metadata Utils)
- General lives at /settings/general so the hub can own /settings
- Desktop /settings redirects to general.
- The settings mobile drawer and its context are gone
  - Back button on the app bar owns navigation

### Other app bar buttons
- Admins get a Settings icon button on the app bar on library pages
   - mobile goes to Settings hub, desktop to General
-  Upload stays permission-based (not admin-only), matching Vue
   - it will move to the library side rail in one of the next PRs (per #173)
- Chromecast is still on the app-bar. Should we move it as well? 
- Upload and Settings are removed from the user menu
   - it will eventually only contain user actions
- audiobookshelf logo is not displayed on mobile
   - but I added it at the top of the mobile library siderail drawer
   - we can consider returning a smaller version of it to allow quick navigation to the current library homepage.

### Other changes:
- **Empty `/library` home** — Replaces the redirect-only route with a real landing page when no libraries exist (admin onboarding CTA vs. no-access message), and wires navigation fallbacks to `/library`.
  - I think this is a better onboarding experience than throwing the admin directly into the settings pages.
- **Password field hydration** — Suppresses hydration warnings on `TextInput` password fields when password managers inject attributes before React hydrates.
- **Stale OIDC cookie cleanup** — Clears leftover `auth_method` / `openid_id_token` on session reset and logout so DB wipes and local login don’t break logout when OpenID isn’t configured.


modarah:
I'm still checking it and noticed some small stuff like when hovering on account dropdown (admin), the hover on the whole header gets triggered
grafik

The search has a fixed width which causes the other elements like account dropdown to be pushed outside of view
grafik

**I very much like how this looks on mobile!**

mikiher:
> I'm still checking it and noticed some small stuff like when hovering on account dropdown (admin), the hover on the whole header gets triggered 

Sorry, I'm not sure I follow? Can you explain the issue better?
 
> The search has a fixed width which causes the other elements like account dropdown to be pushed outside of view 

Thanks for reporting - I'll fix that.

modarah:
> Sorry, I'm not sure I follow? Can you explain the issue better?
> 
Take a look at the AppBar/Header background color in the first screenshot, it's lighter than the second. For some reason it was changing color when hovering on the account. But ignore me here because I just checked again and the issue is not happening anymore, sorry. 😅

mikiher:
> > The search has a fixed width which causes the other elements like account dropdown to be pushed outside of view
> 
> Thanks for reporting - I'll fix that.

This is fixed now.

advplyr:
I think changing the flow is unnecessary because there are dozens of guides and walkthroughs showing new users how to get started. But I think having to check the user type every time calling `getUserDefaultUrlPath` was unnecessary, so the compromise is only on the first login after init will they go to the library settings page.

advplyr:
Definite improvement. Thanks!

PR #183, opened 2026-08-10
Fix #160: narrator rename/remove confirmation dialogs

### LINKED ISSUE #160: Narrator remove/rename confirm dialogs should use ConfirmDialog
On the narrators page, rename and remove confirmation dialogs use a non-standard `Modal`. They should be using `ConfirmDialog`.

Div-47:
## Description

Fixes #160

Updated the shared `EditList` component to use the standard `ConfirmDialog` component instead of the non-standard `Modal` for rename and remove confirmation dialogs.

Since `EditList` is shared across multiple pages, this also ensures that confirmation dialogs on other pages using the component follow the same standard UI pattern.

### Changes

* Replaced the rename confirmation `Modal` with `ConfirmDialog`
* Replaced the remove confirmation `Modal` with `ConfirmDialog`
* Applied the change at the shared `EditList` component level
* Preserved the existing rename/remove behaviour and actions

### Testing

* Verified narrator rename confirmation
* Verified narrator remove confirmation
* Verified rename and remove flows on other pages using `EditList`
* Verified both confirm and cancel actions work as expected
* Verified existing functionality remains unchanged

@mikiher

### DIFF
--- a/src/components/ui/EditList.tsx
+++ b/src/components/ui/EditList.tsx
@@ -3,10 +3,10 @@
 import { useTypeSafeTranslations } from '@/hooks/useTypeSafeTranslations'
 import { TranslationKey } from '@/types/translations'
 import { Fragment, useEffect, useRef, useState } from 'react'
-import Modal from '../modals/Modal'
 import Btn from './Btn'
 import IconBtn from './IconBtn'
 import TextInput from './TextInput'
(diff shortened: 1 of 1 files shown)


mikiher:
Hey, sorry, this got obsoleted by PR #179 which also fixes the issue on the way.

Div-47:
Okay, no worries @mikiher . Could you assign some more issue to me ?

PR #199, opened 2026-08-12
feat(player): fullscreen player

### LINKED ISSUE #187: feat(player): fullscreen player
Adds an immersive fullscreen player opened from the expand button in the mini player, bringing the web client closer to the mobile app now-playing screen.

**Original PR:** advplyr/audiobookshelf#5385  
**Related requests:** advplyr/audiobookshelf#4062, advplyr/audiobookshelf#2569, advplyr/audiobookshelf#3345

Considered important for mobile UX. @sokol-matija — if you are interested, would you like to port this to the React client?

sokol-matija:
Closes #187.

Adds a fullscreen player to the mini player, bringing the web client closer to the mobile app now playing screen.

Rewritten after @mikiher's review. Mobile is the primary target now, and the reference throughout is `advplyr/audiobookshelf-app`'s now playing screen rather than the earlier Vue attempt in advplyr/audiobookshelf#5385, which was never reviewed.

### Structure

Fullscreen is split into `PlayerFullscreenMobileLayout` and `PlayerFullscreenDesktopLayout` behind a media query — the same split the mini player already uses — instead of one layout with `lg:` modifiers. That single decision was behind most of the review points. The desktop layout is only chosen when the primary input can hover, so a tablet wide enough to clear `lg` still gets the touch layout.

Both layouts render the mini player's own `PlayerSecondaryToolbar`, so fullscreen and mini have identical capability by construction rather than by my remembering to keep them in sync.

### Review points

| Feedback | Change |
| --- | --- |
| Chapters button alone on the bottom row | Row replaced by the shared secondary toolbar |
| Chapters icon differs from the mini player | `queue_music` removed; `format_list_bulleted` everywhere |
| Chapter list has z-order issues with the button column | Both the panel and the column are gone |
| Two different chapter UIs, and the panel's palette is foreign | One `ChaptersModal`, shared with the mini player |
| None of the buttons in the button column work | Rail deleted; nothing in the touch path depends on hover |
| Chapter name appears twice | Bespoke header removed; `PlayerTrackBar` owns it |
| Play and skip should be much larger | New `hero` size: play 40px → 64px, jumps → 48px, chapter skip → 44px |
| `open_in_full` is a poor affordance | Removed. The artwork is the target, as in the app |
| The ring around the artwork invites accidental seeks | Removed, with its geometry module and tests |
| Controls missing versus the mini player | Shared toolbar, plus a close button |
| Show both chapter and book bars when "use chapter track" is on | `PlayerTrackBar` gains a `scope` prop; `PlayerFullscreenTrackStack` stacks them |

### Why those three were deleted rather than adjusted

**The control rail** was `pointer-events-none`, re-enabled only by `group-hover`. Tailwind v4 compiles that behind an inner `@media (hover: hover)`, so on a touch device the rule restoring pointer events does not exist in the stylesheet at all. The controls rendered fully opaque and were inert for the lifetime of the player. It is invisible from a desktop browser, because a mouse always hovers, and Chrome's device emulation still fires hover — only a real touch device reproduces it. That is on me; I had not tested on hardware.

**The ring** had a 26px hit band whose centreline sat 3px outside the artwork, so the outer 10px of the cover on all four sides was a seek surface. `pointerup` committed unconditionally, with no drag threshold. The 0%/100% seam fell at bottom centre of the artwork, where a thumb rests, and the nearest-sample lookup had no maximum distance, so a touch a couple of pixels either side of it jumped to the start or the end of the book.

**The chapter panel** was a flex sibling of the artwork at a fixed `w-96`. On a 390px viewport it squeezed the artwork stage to about 6px. It also carried its own palette — the current row used `bg-warning`, which means sleep timer or caution everywhere else in the app.

### Two pre-existing `PlayerTrackBar` defects fixed

Not introduced here, but this change mounts the component up to three times at once and makes it the only seek surface on mobile, so leaving them would have been worse than before:

- It handled `onMouseMove` / `onClick` only. Touch synthesises `mousemove` before `click` but never `mouseleave`, so one tap set `isHovering` permanently and left the timestamp pill and cursor line painted over the player. There was also no drag to scrub on touch at all.
- In chapter scope with no chapter under the playhead — end of book, or a gap between chapters — the bar spanned a zero-length range, so `time = base + perc * 0` and a tap called `seek(0)`, losing the listening position with no undo.

It now uses pointer events with the fraction clamped and read from the live rect rather than a width measured only on window resize, falls back to book scope when there is no current chapter, and carries `role="slider"` with keyboard seeking.

### Also

- Title and author links now leave fullscreen instead of navigating behind the overlay.
- `ChaptersModal` timestamps are scaled by playback rate, matching every other timestamp in the player.
- Volume slider got `touch-action: none`; without it the browser can claim a vertical drag as a scroll and fire `pointercancel` mid-drag.
- The playback rate popover dismisses on `pointerdown` rather than `mousedown`.
- Artwork sizing is a pure function in `lib/player/coverFit.ts` with unit tests, replacing the ring geometry tests. It measures the content box rather than `clientWidth`, and both axes are hard limits, so a phone in landscape shrinks the artwork instead of clipping it.
- `role="dialog"` without `aria-modal`, deliberately: the popovers opened from here portal to `document.body`, so a focus trap would make them unreachable by keyboard. Focus moves in on open and is restored on close.

### Known gaps, not addressed here

- **Volume and mute are no-ops on iOS.** `LocalAudioPlayer` sets `HTMLMediaElement.volume`, which iOS ignores, and mute is implemented as volume 0 rather than `.muted`, which iOS does honour. The UI reports success either way. This predates the PR and lives in the player engine; happy to do it separately.
- Landscape does not split cover and controls the way the app does.
- The background is a blurred cover rather than the app's extracted average colour. `useCoverAccentColor` already exists if that is preferred.
- The playback rate popover flashes at the top-left corner for one frame on open — `VolumeControl` guards against this and the rate widget does not.

### Verification

`tsc --noEmit`, `eslint`, `prettier --check` and `next build` all clean. Component suite: 639 tests, 618 passing; the 4 failing specs (`AppBarNav`, `LibrariesDropdown`, `MultiSelect`, `Tooltip`) fail identically on `master` — I re-ran them against a stashed tree to confirm.

Net −401 lines.

Tested on a physical phone this time, not emulation.


mikiher:
There are many issues with this (before looking at a single line of code).

This should focus on the mobile full-screen view (desktop is nice to have, but mobile is really needed). You should polish it. 

Here are some things I noticed on mobile:
   - The chapters button is the sole control on the bottom row - it looks misplaced.
   - You changed the chapters button icon from how it appears in the minified view. why?
   - The chapters list has z-order issues with the button column
   - The chapters list coming from the right is a nice touch, but its styling and color palette is completely different from everything else in the app. Also, we now have one kind of chapters list in the mini player and another kind in the full screen player? that's confusing.
   - None of the buttons in the button column work
   - The chapter name appears in two places
   - The play and skip buttons should be much larger - larger controls are one of the big points of full screen player.
   - The open_in_full button is not a good icon. Most audiobooks/podcast players just use the preview image (and sometimes also the title space) to open full screen - again, this should be a big touch target, so people don't have to fumble to hit it.
   - I don't get the sophisticated track bar around the cover image. It's logic is unclear, and it is prone to accidental touches that might take you away from you current position and really annoy you.
   - Some buttons that appear in the mini player are gone from full screen player. I understand the choice to make some of the controls less visible on full screen (if it was indeed a design choice rather than an oversite), but you have to access them somehow without moving back to mini player mode.
   - There's more vertical space in full screen - if "Use Chapter track" is enabled, you can display both chapter and book track bars (audiobookshelf-app does)

There are also some issues with the desktop full screen, but let's focus first on resolving mobile.

sokol-matija:
Sorry — you were right, and you shouldn't have had to write that list. I built this against a desktop browser and never put it on a phone, which is exactly why every one of those problems survived to your review.

The dead controls are the clearest example. The control rail was `pointer-events-none`, re-enabled only by `group-hover`, and Tailwind compiles that behind `@media (hover: hover)`. On touch, the rule that restores pointer events does not exist in the stylesheet — so the buttons rendered fully opaque and did nothing. It cannot reproduce on desktop, because a mouse always hovers, and Chrome's device emulation still fires hover. Only real hardware catches it, and I hadn't used any.

**What's fixed** — all eleven points, mapped one by one in the description:

- Control rail deleted. Both layouts now use the mini player's `PlayerSecondaryToolbar`, so fullscreen and mini can't drift apart, and everything reachable in one is reachable in the other. Added a close button too — you previously had to minimize before you could stop playback.
- Ring deleted, with its geometry module and tests. It was worse than it looked: the 0%/100% seam sat at bottom centre of the artwork with no distance guard, so a thumb resting there could jump to the start or the end of the book.
- Chapter panel deleted. Chapters open the same `ChaptersModal` as the mini player — one chapters UI, app palette, no z-order problem. Icon is `format_list_bulleted` again.
- Play 40px → 64px, every target clears 44px.
- `open_in_full` gone; the artwork is the target.
- Duplicate chapter name gone; both track bars stack when "use chapter track" is on.
- Fullscreen is split into real mobile and desktop layouts, the same split the mini player uses. That was the root cause of most of the list.

I also fixed two pre-existing `PlayerTrackBar` bugs, since this change makes it the only seek surface on mobile: a tap left the hover readout stuck permanently, and in chapter scope at the end of a book a tap called `seek(0)` and lost the position.

Known gaps are at the bottom of the description. The one worth your attention: volume and mute are no-ops on iOS — `LocalAudioPlayer` sets `.volume`, which iOS ignores, and mute goes through volume 0 instead of `.muted`, which iOS does honour. Predates this PR; happy to take it separately.

Net −401 lines. Tested on a physical phone this time.

mikiher:
Thanks, this looks much better.

A few more issues:

- Back button should close full-screen, not navigate back from the page below it.
- When pressing Escape over popovers (like the speed rate popover or the volume knob), it closes the player instead of just closing the popover
   - I think this pre-dated your changes, but please fix this anyway
- There are issues with mobile landscape mode:
   - The top part (book details) is clipped
   - The bottom part takes too much real estate
   - I think full screen player should generally get an optimized landscape layout (maybe the mini player as well)
- There's a redundancy between the volume popover and the volume HUD (beside the cover on full-screen)
   - when navigating with keys, only HUD appears
   - when clicking the the volume button, both display
   - I think the HUD is nice, but the duplication is weird and confusing.
- Full-screen tabbing can move to tab targets below the full-screen. In full screen player, the background should be inert.
  - Try opening full screen player and then Shift-tab a number fo times - you'll see tooltip from the mini player controls that get focused
   
Nits:
- Both sliders have the same aria-label. Please distinguish.
- Cover size state starts at MIN_COVER_WIDTH. Hide or delay paint until the first measure.
- Queue button height is non-standard,
   - probably pre-dates your change. please fix anyway.

Also, please resolve conflicts with master

mikiher:
Oh, and one more request (optional) - it would be great if you could implement nice transition animations between mini and full-screen and back (and also between full-screen potrait and landscape), similar to what's implemented in audiobookshelf-app. Not strictly necessary, but very nice to have.

sokol-matija:
Thanks for the detailed feedback — really appreciate it. I'm working through all of it now: back button closing full-screen, Escape scoped to popovers, an optimized landscape layout, removing the volume HUD/popover duplication, making the background inert, plus the nits (distinct slider labels, cover measurement, queue button height). I'll take a look at the mini ↔ full-screen transitions from audiobookshelf-app as well.

If you have any more thoughts on the UX while I'm in there, please don't hesitate to share — happy to take any input that makes it feel better.

sokol-matija:
All of the above is fixed. While I was in there I also brought three things back — but desktop-only, and each one addresses the specific objection you raised rather than ignoring it. Happy to drop any of them if you still disagree.

**Progress ring around the artwork.** Your objection was accidental touches taking you away from your position. That was correct, and it was a touch problem: the grab target is a 26px transparent stroke, which is a thumb-sized hazard and a cursor-sized convenience. It is now rendered only by the desktop layout, which is already gated on the primary input being able to hover — a phone or a tablet never sees it, whatever the width. On desktop it is the only way to scrub without leaving the artwork, and the track bar below the transport controls remains the accessible seek surface (the ring is `aria-hidden`).

**Controls on a rail beside the artwork.** Your objection was that none of them worked. That was a real bug, not a design choice: the rail was `pointer-events-none`, re-enabled by `group-hover`, and Tailwind compiles that behind `@media (hover: hover)` — so on touch the rule that restores pointer events did not exist in the stylesheet. The rail now animates opacity and transform only; `pointer-events` is never touched, so anything painted is clickable. It also reveals on `focus-within`, so it is reachable by keyboard. Desktop only — the touch layouts keep the always-visible `PlayerSecondaryToolbar`. Nothing reachable in the mini player is missing in either: the rail carries volume, speed, sleep and bookmarks, the bottom bar carries chapters, queue and settings.

**Chapters docked beside the artwork.** Your objection was two things — a palette that matched nothing else, and one kind of chapters list in the mini player and another in full-screen. Both are gone: I extracted the modal's rows into a shared `ChapterList`, and the modal and the docked panel now render the same component. Same rows, same tokens, same behaviour. Only the container differs, and only at desktop width, where the artwork and the list both fit and covering the artwork to pick a chapter is the worse trade.

Two smaller things, both ported from the Vue client rather than invented here:

- The play button is white in the player, matching `whitePlayButton` in `PlayerPlaybackControls.vue`, which the Vue full-screen player passes. The accent competes with the artwork behind it.
- The full-screen track bars use the Vue client's `rounded` variant (`h-2.5 rounded-full`).

And the desktop bottom bar is now the same three-column grid the Vue full-screen player uses — chapters, queue and settings sit level with the transport controls instead of adding a row, which is what made it a third taller than it needed to be.

sokol-matija:
Adding the rest of what landed, so none of it comes as a surprise in the diff.

**Your list**

- Back button closes full-screen. Opening pushes a throwaway history entry; closing from the UI pops it, so the button is never left needing a second press. Following a link out of the player skips the pop, or it would undo that navigation.
- Escape closes an open popover, not the player. The popovers portal to `document.body`, so nothing in the tree could see them — they now publish their open state to a small store the global hotkey handler checks before acting on Escape. `VolumeControl` had no Escape handler at all.
- Background is inert. The overlay is portalled to `document.body` so every other body child can be marked `inert`. Anything portalled *after* that — popovers, tooltips, modals opened from inside — is deliberately left alone, which is what a plain focus trap gets wrong: those live outside the overlay's subtree but belong to it.
- Mobile landscape has its own side-by-side layout, no separate bottom bar. Chosen by `(orientation: landscape) and (max-height: 600px)`.
- Volume readout and volume popover no longer both appear.
- Nits: the two stacked sliders have distinct `aria-label`s, the artwork is not painted until the stage has been measured once (`useFittedCoverWidth` returns `null` until then), and the queue button gets the same box as its neighbours.
- Conflicts with master are resolved — this branch is merged up to `1c9410af`. I took your `PlaceholderBookmarkNote` and `AriaLabelJumpForwardWithDuration` / `AriaLabelJumpBackwardWithDuration` changes across into the extracted panels rather than dropping them.

**Also in, and worth knowing about before you read the diff**

- Transitions between mini and full-screen, and across the portrait/landscape swap — the optional one you asked for. All of it honours `prefers-reduced-motion`.
- Keyboard shortcuts sheet on `?`. None of the hotkeys were discoverable before. While I was there I ported two bindings this client was missing: `L` for chapters, and `.` / `,` for the playback rate.
- Playback rate pill, shown only while the rate is off 1x — a permanent "1.0x" is noise, the interesting state is the one you forgot you left on.
- White play button and pill-shaped track bars, both ported rather than invented: the Vue client already passes `whitePlayButton` and `rounded` in its full-screen player.
- Player settings for the ring, the queue button, the corner buttons, AMOLED surfaces, bookmark and volume placement, auto-open on play, and legacy dialogs for the sleep timer and bookmarks. Anything you disagree with can be defaulted the other way rather than argued about.

Nothing in the desktop path can regress the mobile path you reviewed — the layout is chosen by whether the primary input can hover, not by width, so a wide tablet still gets the touch layout. Mobile's structure is unchanged: artwork, title, track stack, transport, `PlayerSecondaryToolbar`.

mikiher:
> While I was in there I also brought three things back — but desktop-only, and each one addresses the specific objection you raised rather than ignoring it. Happy to drop any of them if you still disagree.

Sorry, but no. This PR is big enough as it is. 

Please revert all of these and we'll be happy to review them ***one by one*** in future PRs (you also accidentally reverted some of your initial fixes, e.g. using the wrong icon for chapters, so please be careful).

Please focus on fixing what I asked for (and any other issues you found), rather than on adding new functionality.

Also generally, the sentiment here is that settings are already complex enough - so cramming many new display options in the settings will be frowned upon.

sokol-matija:
Reverted. This branch is now your list and nothing else.

The chapters icon is `format_list_bulleted` again — that was a regression, thanks for catching it.

**What's left**

- Back button closes full-screen instead of navigating away from the page underneath.
- Escape closes an open popover rather than the player behind it.
- The background is inert while full-screen is open.
- Mobile landscape has its own side-by-side layout, no separate bottom bar.
- The volume readout and the volume popover no longer both appear.
- The two stacked sliders have distinct `aria-label`s.
- The artwork is not painted until the stage has been measured once.
- The queue button gets the same box as its neighbours.
- Enter/exit and portrait↔landscape transitions, the optional one you asked for. Honours `prefers-reduced-motion`.
- Merged up to master; conflicts resolved. I kept your `PlaceholderBookmarkNote` and `AriaLabelJump*WithDuration` changes rather than dropping them.

**What went out**

The scrub ring and its geometry module, the hover rail, the docked chapters and bookmarks panels, the sleep timer popover, the shortcuts sheet, the rate pill, the white play button, the pill-shaped track bars, the three-column bottom bar, and every display setting. Net −1941 lines against the previous push.

One thing I should correct from my last comment, since it may have been my fault for not saying where they lived: those settings were all in the player settings modal — the one that holds "use chapter track" and the two jump amounts — not in the app or server settings pages. That doesn't make more toggles right, and they're gone either way. I just didn't want you weighing this against the wrong surface.

I'll follow this with a short clip of what came out, so you can tell me which are worth your review time and which to just drop — I'd rather bin three of them now than send you three PRs you didn't want. Anything worth pursuing I'll send one at a time, each with a fixed sensible default and no setting attached, which removes the complexity objection entirely and keeps each one small enough to judge in a sitting. If the answer is none of them, that's a fine answer and this PR stands on its own.

sokol-matija:
Here's the clip — about a minute, desktop full-screen, no audio.

https://github.com/user-attachments/assets/f6f4da43-962f-4eae-b52e-2fc4c03d8b58

In order: the scrub ring on the artwork, the controls rail beside it, chapters docked instead of covering the cover, the sleep timer as a popover on its own button, the rate pill, and the ? shortcuts sheet.

No need to write much — a list of which ones you'd take a PR for is enough, and "none" is a perfectly good answer. Each one I do send will have a fixed default and no setting.

sokol-matija:
Mobile too, 
  
https://github.com/user-attachments/assets/04dc72ec-26b9-4705-9eb8-abe924bed322

  Chapters and bookmarks as a bottom sheet instead of a modal — drag down or tap outside to
  dismiss. Same `ChapterList` the modal renders, different container, and on touch it
  replaces the modal rather than adding a surface. One chapters UI, not two.

  Planning to send this one on its own next, off the current branch.

mikiher:
I fixed a couple of additional issues and polished the player controls and landscape layout.
@advplyr I'm passing this to you for review and final approval. This looks good from my perspective

advplyr:
Sorry for the delay on this, I've been trying to think about what is off. I think we have to do re-do this and maybe strip down the media player and think about it with this feature in mind.

Ideally when you expand to fullscreen all of the existing components like the cover, title, track bar, play button, etc., remain the same components. With CSS those components are scaled and moved into position as the media player expands upwards. It shouldn't look like a separate component is being rendered.

You can see this in the mobile app. We can probably do it even better because there is some flickering that occurs in the mobile app.

advplyr:
I know it may seem nitpicky but I think this part of the design is one of the most important that we get right.

Side note so we don't forget: We also support having the e-reader open while the (minimized) media player is open. For users who read along and listen or for audiobooks that have supplementary PDFs, etc.

mikiher:
I'm going to close this right now as it's going to be superseded by #273. Thanks for working on this!

PR #200, opened 2026-08-13
Link listening session modal title to the library item page

### LINKED ISSUE #195: feat: link listening session modal title to library item
In Settings → Listening Sessions, make the item title in the session detail modal a link to `/item/{libraryItemId}` instead of static text.

**Original PR:** advplyr/audiobookshelf#5410

Small UX improvement — every other title in the UI is clickable. @drewmerc302 — would you like to port this to the React client?

drewmerc302:
## Summary

Port of advplyr/audiobookshelf#5410 to the React client, as requested in #195.

In Settings → Listening Sessions, the session detail modal showed the item title as static text. This renders it as a link to the item page instead, with underline on hover. Clicking the link closes the modal before navigating.

One deviation from the issue text: #195 suggests linking to `/item/{libraryItemId}`, but this client has no bare `/item/` route. The link targets `/library/{libraryId}/item/{libraryItemId}` instead, matching the existing pattern in `PodcastDownloadQueueTable.tsx`. Since `libraryId` is optional on a session (the modal already guards it in the details rows), the title falls back to plain text when either id is missing.

## Testing

- `tsc --noEmit` and `eslint` pass.
- Verified manually against a local server: opened a closed session's detail modal, clicked the title, modal closed and landed on the item page; sessions lacking ids render the title as plain text.

Closes #195

https://claude.ai/code/session_01HMVbHV6BGmXUdG7XDLrAmr

### DIFF
--- a/src/app/(main)/library/[library]/LibraryEmptyState.tsx
+++ b/src/app/(main)/library/[library]/LibraryEmptyState.tsx
@@ -1,4 +1,5 @@
 import Btn from '@/components/ui/Btn'
+import PageMessage from '@/components/ui/PageMessage'
 import { useTasks } from '@/contexts/TasksContext'
 import { useTypeSafeTranslations } from '@/hooks/useTypeSafeTranslations'
 import { Library } from '@/types/api'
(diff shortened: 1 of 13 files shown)


mikiher:
One failure mode is that listening records may outlive the the items they point to (this is by design), but the addition of the link exposes you to the risk of the item not being found. I realized we didn't yet properly handle this, so I added it.

Thanks!

PR #203, opened 2026-08-14
Add infinite scroll to the podcast Latest page

### LINKED ISSUE #196: feat: infinite scroll on podcast Latest page
Add infinite scroll (with load-more fallback) to the podcast Latest page so users can browse beyond the first 50 episodes. Addresses advplyr/audiobookshelf#3343.

**Original PR:** advplyr/audiobookshelf#5338  
**Notes:** `/api/libraries/:id/recent-episodes` already supports paging but does not return total count, so scroll length cannot be calculated precisely until a minor server change. Client paging infra exists for books/authors/etc.

@wakamex — would you like to port this to the React client?

wakamex:
## Summary

Adds infinite scroll with a load-more fallback to the podcast Latest page, porting the behavior from advplyr/audiobookshelf#5338 to the React client.

The first 50 episodes are fetched during the server render. As the user nears the end of the list, the client fetches the next page through a server action and appends new episodes without duplicates. Episode rows are virtualized with measured variable heights, so only the visible range and a small overscan remain mounted. The More button remains available as an accessible fallback and for retrying after an error.

Review note: 250 of the 620 added lines are Cypress tests. The runtime change is 370 additions and 9 deletions across seven files, including the measured row virtualizer.

Closes #196

## In-depth description

The pagination hook owns the loaded episodes, next page, end-of-list detection, duplicate protection, and concurrent-request guard. The virtualizer measures rendered rows with ResizeObserver, estimates unseen row heights, and positions only the visible range plus three overscan rows. This preserves variable-height titles and descriptions without allowing the mounted React tree to grow with every loaded page.

The load observer uses the nearest scroll container with a 600px prefetch margin rather than assuming the browser viewport.

The current recent-episodes API is offset based and sorts by publication time, so pages can shift if episodes change between requests. This PR uses that existing API and deduplicates by episode ID. A stable server cursor with a deterministic tie-breaker would be the ideal API end state, but it requires a separate server change.

## Testing

- pnpm check
- pnpm build
- Cypress component tests: 7 passing, covering pagination, retries, duplicate handling, concurrent-load protection, IntersectionObserver loading, bounded mounted rows, scrolling virtual ranges, and variable-height measurement
- Local smoke test against a podcast library with more than 7,700 episodes: loaded the second page, kept fewer than 20 episode rows mounted, reached virtual indexes above 50, and restored the first rows after scrolling back to the top

## Screenshots

No screenshot included because the episode layout is unchanged. The visible change is the existing-style More button at the end of the list, followed by automatic loading as it approaches the scroll viewport.

AI assistance: OpenAI Codex was used during implementation and testing. I reviewed the final diff and verified the behavior against my local server.


mikiher:
This code has no virtualization. 

I consider it a must for infinite scroll in a large library, otherwise page memory is unbounded, and rendering performance also degrades as you load more pages. can you please implement?

wakamex:
Thanks, agreed. I added variable-height row virtualization in 368cba84. It measures rendered rows with ResizeObserver and keeps only the visible range plus three overscan rows mounted, so titles and descriptions can keep their natural heights.

I added tests for bounded mounted-row count, scrolling through virtual ranges, and variable-height measurement. The full focused suite now has seven passing tests, and a smoke test against my local library loaded beyond episode 50 and returned to the first rows while keeping the mounted row count bounded.

I also measured retained memory because virtualization only bounds the React tree and DOM, not the loaded episode data. Using the production build with forced garbage collection at each checkpoint:
  - 50 episodes: 30.4 MB used JS heap
  - 1,000 episodes: 47.0 MB
  - 2,500 episodes: 70.9 MB
  - 5,000 episodes: 104.5 MB
  - 7,765 episodes: 144.9 MB

Mounted rows stayed between 5 and 10 after the initial render, so the remaining growth is retained episode data. The increase from 50 to 7,765 episodes was about 114.5 MB.

The endpoint currently repeats the full podcast object on every episode. A compact or normalized response could reduce that substantially. Fully bounded data memory would also require page eviction and backward refetching, which I do not think we should build over the current offset API because inserts and finished-state changes can shift pages.

Would you prefer to keep #203 as the client-side pagination and virtualization change, with the compact cursor API and data eviction as a server follow-up? Or should the server-first design be a prerequisite for this PR?

wakamex:
One extra detail from profiling the response: repeating the full podcast object accounts for only about 11% of the sampled payload. Roughly 62% is overlapping audioTrack and audioFile metadata, and another 19% is episode descriptions. A smaller response containing only the fields needed by this list would therefore save considerably more memory than deduplicating the podcast data alone.

PR #204, opened 2026-08-16
feat(player): apply natural volume scaling

### LINKED ISSUE #190: feat(player): natural volume scaling
Apply a quadratic volume curve so the slider feels more natural across its range (finer control at low volumes). Fixes advplyr/audiobookshelf#4889.

**Original PR:** advplyr/audiobookshelf#4935  
**Decision:** Implement for everyone — no new player setting.

@lerndmina — would you like to port this to the React client (without adding a toggle)?

blue219:
## Brief summary

Applies quadratic volume scaling to local audio playback so the volume slider provides finer control at low volumes.

## Which issue is fixed?

Fixes #190

## In-depth Description

`LocalAudioPlayer` now maps the linear slider value to `volume²` before assigning it to the HTML audio element. This keeps the slider range and persisted values unchanged while producing 25% output at 50% slider position and 100% output at 100%.

The behaviour applies to all local playback without adding a player setting or toggle.

## How have you tested this?

- `node_modules/.bin/eslint.cmd src/lib/player/LocalAudioPlayer.ts cypress/tests/lib/player/LocalAudioPlayer.cy.tsx`
- `node_modules/.bin/cypress.cmd run --component --browser electron --spec cypress/tests/lib/player/LocalAudioPlayer.cy.tsx` (1 passing)
- `git diff upstream/master...HEAD --check`

## Screenshots

Not included: this is an audio-output behaviour change with no visual UI change.

### DIFF
--- a/cypress/tests/lib/player/LocalAudioPlayer.cy.tsx
+++ b/cypress/tests/lib/player/LocalAudioPlayer.cy.tsx
@@ -1,0 +1,29 @@
+import { LocalAudioPlayer } from '@/lib/player/LocalAudioPlayer'
+
+describe('LocalAudioPlayer volume scaling', () => {
+  let player: LocalAudioPlayer
+
+  beforeEach(() => {
+    cy.mount(<div />)
+    cy.then(() => {
+      player = new LocalAudioPlayer()
+    })
+  })
+
+  afterEach(() => {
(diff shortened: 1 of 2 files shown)


mikiher:
Thanks!

PR #206, opened 2026-08-16
Fix chapter track when chapters are unavailable

Neon1201:
## Brief summary

Fixes chapter track behavior when the current audiobook has no chapters.

If "Use chapter track" was enabled, books without chapters could still enter chapter-track mode. Since there was no current chapter, the chapter duration became 0 and clicking the progress bar would always seek to 0:00.
click
click22
du
image

This change:
I changed the player so that books without chapters use the full audiobook track instead of chapter track. The chapter track setting is disabled for those books, but the saved preference is kept, so it becomes active again when switching back to a book with chapters.

## How have you tested this?

I tested two books with chapters and one book without chapters, and switched between them several times to make sure the chapter track setting behaved correctly.

### DIFF
--- a/src/components/player/PlayerModals.tsx
+++ b/src/components/player/PlayerModals.tsx
@@ -42,6 +42,7 @@ export default function PlayerModals({ controls }: PlayerModalsProps) {
       <PlayerSettingsModal
         isOpen={isSettingsModalOpen}
         settings={settings}
+        hasChapters={chapters.length > 0}
         onClose={() => setIsSettingsModalOpen(false)}
         onUpdateSettings={playerHandler.controls.updateSettings}
       />
--- a/src/components/player/PlayerSettingsModal.tsx
+++ b/src/components/player/PlayerSettingsModal.tsx
@@ -9,6 +9,8 @@ import { useTypeSafeTranslations } from '@/hooks/useTypeSafeTranslations'
 interface PlayerSettingsModalProps {
   isOpen: boolean
   settings: PlayerSettings
+  /** Whether the currently playing item has chapters. Does not mutate the saved preference. */
+  hasChapters?: boolean
   onClose: () => void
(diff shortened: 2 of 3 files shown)


mikiher:
I believe this issue is mostly/completely fixed in #199 which is also currently under review and has many other player related changes. 

Can you please check that PR and see if it resolves the issue. If there's anything else to fix, let's wait until #199 is merged and continue from there.

Neon1201:
> I believe this issue is mostly/completely fixed in #199 which is also currently under review and has many other player related changes.
> 
> Can you please check that PR and see if it resolves the issue. If there's anything else to fix, let's wait until #199 is merged and continue from there.

Thanks, I checked #199. It looks like it already fixes the main issue where chapter-track mode can fall back to a zero-length range and seek to 0:00.

One small difference in my PR is that it also disables the “Use chapter track” setting when the current book has no chapters, while keeping the saved preference so it becomes active again for books with chapters.

I’ll wait for #199 to be merged, then test again and see whether that remaining behavior is still worth addressing separately.

mikiher:
> One small difference in my PR is that it also disables the “Use chapter track” setting when the current book has no chapters, while keeping the saved preference so it becomes active again for books with chapters.

I have noticed that. 

I think my preference would be to hide that settings altogether when there are no chapters. This way there would be no confusion about the value and why the setting is disabled.

Neon1201:
That makes sense. My only concern with hiding it completely is that the settings UI would change between books depending on whether chapters are available. I used the disabled state to keep the layout consistent and indicate that the option still exists, but is unavailable for the current book.

That said, I’m happy to follow the project’s preferred behavior.

mikiher:
Since #199 will likely take more time, I submitted a60dd39 which fixes this issue as discussed above.
Thanks for taking the time to send this PR!

PR #207, opened 2026-08-17
Add PWA support

### LINKED ISSUE #198: Implement PWA support
The React client currently has no PWA support. This was raised during testing discussion in #112 — the web client is used heavily on phones (especially iOS), and PWA was overlooked in the migration checklist.

## Goal

Add PWA support to the React client with **Vue parity as the baseline**, while evaluating whether newer PWA practices should be

## Vue reference

The Vue client uses `@nuxtjs/pwa` with installability-focused settings and **explicitly disables offline usage**:

- `display: standalone`, app name/icons, theme/background colors
- iOS meta tags (`apple-mobile-web-app-capable`, status bar style, etc.)
- Workbox with `offline: false`, no precaching, no runtime caching

See `client/nuxt.config.js` in advplyr/audiobookshelf.

## In scope

- Web app manifest (name, icons, `standalone` display, theme colors)
- Service worker setup appropriate for an online-first app (e.g. update prompts / skip waiting if useful)
- iOS home-screen / Add to Home Screen behavior
- Verify installability and standalone launch on mobile browsers

## Out of scope

- **Offline media playback** — intentionally excluded in Vue and should remain out of scope here (major effort; OPFS/quota concerns on iOS)
- Precaching app shell or API responses for offline browsing

## Notes

- Related mobile UX requests: advplyr/audiobookshelf#4062, #2569, #3345 (fullscreen / mobile playback UI — see also #187)
- Implementation approach is open (e.g. `@serwist/next`, `next-pwa`, or manual Workbox) — choose what fits Next.js best and document tradeoff

modarah:
Closes #198

- Web app manifest: added (name, icons, standalone display, theme/background colors), base-path aware for subfolder deploys.
- Icons: added 512px and maskable variants + an opaque iOS home-screen icon.
- Install/iOS metadata: appleWebApp, apple-touch-icon, and themeColor/viewport added in the layout so it installs and launches standalone.
- Minimal service worker
- Excluded sw.js and manifest.webmanifest from auth redirects in the proxy matcher; Cache-Control: no-cache on /sw.js in next.config.ts.
- Cypress tests for the manifest builder and SW registration.

I didn't use any dependencies because any advantages we will get are not yet needed like caching such.

I tested it on iPhone 13 mini. I needed to set up an HTTPS reverse proxy so the service worker gets served.

IMG_9431

<br>

IMG_9430


mikiher:
ROUTER_BASE_PATH is an API server setting that should not be used on the React client excpet in special circumstances.

NextJS basePath support is not yet implemented, so I changed ROUTER_BASE_PATH to NEXT_PUBLIC_BASE_PATH, which is currently unset, so base is always empty. When we implement basePath, this will server as a marker for us to re-test this with a non-empty base.

I also fixed a few nits, mainly changing the app name to use a capital A (Vue parity)

Other than that, this looks good, and I also tested it on my Android phone and iPad. 
Both install the PWA correctly and it is displayed properly without showing browser chrome.

Thanks!

PR #210, opened 2026-08-18
Chapters Edit Section

mikiher:
_**Note: this is for functionality review, desktop only (mogile not well-thought yet). Code is not ready for review yet.**_

This puts Chapters editing as a library item edit metadata modal section, instead of `/chapters` page. 

There are two modes (View, Manual Edit), plus bulk transforms that stage a replacement. Save is a separate step.

### View vs Manual Edit

Opens in **View**: read-only list (titles, times, durations, play). Toolbar: **Lookup**, **Set from tracks**, and **Manual Edit** (or **Add Chapters** if empty).

**Manual Edit** is row-level: titles, start times, insert/remove, title-pattern add, remove all, shift the editor list, seconds display. Enter with **Manual Edit** / **Add Chapters**; leave with **Done Editing**. 

Lookup / Set from tracks are hidden here so they cannot collide with in-place edits.

Typical flow: View → bulk transform → Apply (still View) → Manual Edit if needed → Done Editing → Save.

### Bulk vs in-place

**Lookup** and **Set from tracks** open a Current vs New preview. Current is the working list (including unsaved edits). New is staged. **Apply** copies New into the editor; it does not save. Discard and Save still apply to the editor list.

**Manual Edit** changes the working list in place. Shift there applies immediately (still unsaved).

### Before Apply

Tune New without committing:

- **Shift (seconds)** — offset New starts (first chapter stays at `0:00`)
- **Map titles** (Lookup) — keep current starts, take Audible titles
- **Remove branding** (Lookup) — strip intro/outro branding

Yellow/red New rows: start or end past the audiobook duration.

### Dirty coloring

Changed fields use `text-info` vs the last **saved** chapters (not last Apply):

- Preview: New start/title vs Current
- Table: start, title, duration per field; new rows fully dirty; shifting a start also dirties the previous duration

Unchanged fields stay default color. Highlights remain until Save or Discard.

### Locked chapters removed

The old editor could lock rows. That UI is gone since I found it not be very useful, but I'll bring it back if you think it makes sense.


advplyr:
I think we can make something like this work but we have to find a way to simplify it more. All the buttons and sections are confusing.

I don't think the "Done Editing" and toggling between them is intuitive enough.

Maybe there is some way to clearly section these editing modes

mikiher:
OK, that is fair (this is why I called it version 0).

I'll try to use less popping sections, less buttons, no view/edit modes (only edit). Stay tuned.

mikiher:
OK, here's my next attempt. I believe it's simpler and less cumbersome than the previous. PTAL again.

**User-facing changes**

- **Chapters edit UI** 
   - Single-screen modal
   - Lookup + shift + Set from Tracks up top
   - Single editable table for all views
   - Discard/Save at the bottom (no separate apply step). 
   - Removed space consuming +/- buttons.
   - No seconds checkbox (seems not useful)
   - Inline bulk add chapters (no modal)
- **Smarter Lookup / Set from Tracks**
   - Incoming chapters are matched to **saved** ones by start time and title match. 
   - Matched rows keep identity for dirty highlighting
   - ±1s start drift doesn’t count as dirty.
- **Keep existing start times**  (instead of the not-very-useful Map Chapter Titles)
   - Uses the found chapter list as basis
      - restores your starts on matches
      - keeps unmatched found chapters
      - drops unmatched saved ones.
- **Lookup errors**
   - Failures show as toasts, not text under the ASIN field.
- **Shift** 
   - Negative amounts work again, with number spinners still available.

Again, does not cover mobile yet, and please don't review code for now, just functionality.

**_Note: the Lookup and Set From Tracks results contain text below the start and title fields - those are just for debug and will be removed later. They're meant to show comparison between the old and new fields, and the matching cost score (higher is worse)_**

advplyr:
Nice job, this is much more clear. I would love to simplify it even further where we can but this covers everything better than the existing page I think.

We will also highlight this to testers for feedback.

mikiher:
@advplyr, this is now ready for final review.

In addition to much polish, this now has a mobile layout - I don't think it's the best, but this is what I could achieve while keeping the existing functionality - it's working, and I think that's good enough for now. We should revisit this in the future.

src/lib/chapters/chapterMatching.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export interface ChapterMatchOptions {
+  maxStartDeltaSec: number
+  maxCost: number
+  startWeight: number
+  titleWeight: number
+}
+
+export const DEFAULT_CHAPTER_MATCH_OPTIONS: ChapterMatchOptions = {
+  maxStartDeltaSec: 60,
```
advplyr: I'm thinking this might be a bit loose. Do you know if we are using 60 second delta in the Vue app?

advplyr:
Found a few issues:

- If you edit a chapter title and keep the title input focused when pressing "Discard Changes" button, the change is not actually discarded.
- This may be related to the above. It is possible to have the chapter title be changed, but the buttons are disabled. And vice versa, to have no changes and the buttons are enabled. The dirty state is buggy. You can test by editing a chapter title, blurring the input, changing the chapter title again and blurring the input.
- I think the "Set from tracks" and ASIN search should show the same "You have unsaved chapter changes." prompt. Especially the set from tracks because I could see someone pressing that by accident.
- The next/prev buttons do not trigger the "You have unsaved chapter changes." prompt

src/lib/chapters/chapterMatching.ts
mikiher: There's no equivalent algorithm in Vue - this is new code. 
I think we can safely drop it to 45, but I've seen cases when a 30 seconds delta was legit.

mikiher:
All issues should now be fixed. PTAL again.

mikiher:
Btw, I believe the dirty mechanism and "You have unsaved changes" dialog could benefit the Details (and perhaps the Match) section as well. Should we implement it there as well?

advplyr:
> Btw, I believe the dirty mechanism and "You have unsaved changes" dialog could benefit the Details (and perhaps the Match) section as well. Should we implement it there as well?

Yeah I think we should

advplyr:
This is working great for me now. I decided to leave a lot of the unused i18n strings for now. We can do a pass removing unused strings later.
I think having this in the modal is much better than using a standalone page. Thanks for figuring all that out

PR #213, opened 2026-08-19
Update player chapters when the playing item is edited

### LINKED ISSUE #211: Player chapter data does not update after using Edit Chapters during playback
Playing a book and then editing its chapters does not update the player. View Chapters still lists the old chapters until the player is closed and the book is played again.

Steps:
1. Start playing a book
2. Use Edit Chapters on that book and save
3. Open View Chapters in the player

Expected: the updated chapter data is listed

Actual: The previous chapter data from when playback started is listed. Closing the player and playing the book again shows the
updates.

hypnotoad08:
Fixes #211

Chapters were only set when a playback session started, so chapter edits did not reach a book that was already playing. 

Updated usePlayerHandler - it now listens for item_updated so it refreshes chapters for the item currently playing and calls syncChapterNav.

Changed syncChapterNav to compare chapter objects by reference instead of by start time. The previous comparison would have kept the previous chapter object when only the chapter title changed, so the current, next, and previous chapter objects held stale titles after an edit.

Testing:
- Started a book, edited its chapters, and confirmed View Chapters lists the updated data without closing the player
- Confirmed the current chapter title in the player updates after renaming the chapter title


mikiher:
Works nicely. Also added an items_updated handler, and an optimization not to update chapters if they remain unchanged.

Thanks!

PR #216, opened 2026-08-21
Update search results when collections or playlists change

### LINKED ISSUE #215: Search results do not reflect updates to collections or playlists
Creating, renaming, or deleting a collection or playlist does not affect search results until something triggers an update. Navigating within the app won't resolve the issue.

Refreshing the page, logging out and back in, or switching libraries will show the updated search results.

**Steps to reproduce**

A search is required before creating or modifying the collection or playlist to reproduce the issue.

**New collection or playlist:**

Search for a book or series.
Create a collection or playlist.
Search for the new collection/playlist.

**Expected:** the new collection/playlist appears in the search results.
**Actual:** the new collection/playlist does not appear in the search results.

**Renamed collection or playlist:**

Search for a book or series.
Rename a collection or playlist.
Search for the new name.

**Expected:** the collection/playlist appears in the search results under its new name.
**Actual:** the collection/playlist does not appear under its new name, and searching the old name returns it.

**Deleted collection or playlist:**

Search for a book or series.
Delete a collection or playlist.
Search for the deleted collection/playlist.

**Expected:** the collection/playlist appears in the search results.
**Actual:** the deleted collection/playlist still appears in the search results, and clicking on it results in a 404.

hypnotoad08:
Fixes #215

The collections and playlists used for search results are currently fetched once per library and cached, then filtered client-side, but nothing invalidates that cache when they change. 

A collection or playlist created, renamed, or deleted after a user's first search in a session is missing or stale in the search results until the page is refreshed or the user logs back in. Switching libraries also clears it, so on a single-library server nothing clears the cache during a logged-in session.

Updated useLibrarySearch - it now listens for collection_added, collection_updated, collection_removed, and the playlist equivalents, and clears the fetched flag so the next search refetches and updates the cache.

**Note:** While testing the rename flow with this fix, I noticed two identical search requests were being sent for a single search. The search caches were stored in state, so refetching them changed the handleSearch reference, which caused the debounce effect in GlobalSearchInput to re-run and triggered a second search for the same text. I changed the caches to refs to resolve this issue, as the caches were never rendered.

**Testing:**

- Renamed a collection and a playlist and confirmed they appear in the search results under the new name without refreshing the page

- Created and deleted collections and playlists and confirmed the results update on the next search

- Confirmed only one search request is sent per search in the network tab and the dev server log


mikiher:
- broke the debounce dependency on handleSearch by using React 19 useEffectEvent
- Split the playlist and collection handlers (since they're independent)
- Fixed a couple of nits

Thanks for reporting the issue and fixing it!

PR #218, opened 2026-08-23
Improve item page table layout on mobile

### LINKED ISSUE #139: Fix item page tables mobile layout
It looks like the Chapters, Audio Tracks and other tables on the item page are overflowing horizontally on mobile. This should be fixed not to overflow.

mikiher:
This fixes the `redesign action buttons on book page tables` action item on #112 and also fixes #139

**Notable changes**
- Section actions (edit/add chapters, manage tracks, full path) use compact icon buttons so headers stay usable on small screens
- Chapters, tracks, and files tables fit narrow (~360px) widths without horizontal scroll; relative paths wrap and row menus stay visible
- Chapter Id and track `#` hide on the smallest screens; duration stays visible
- Full path toggle remains desktop-only; mobile always shows relative paths

### DIFF
--- a/src/components/ui/DataTable.tsx
+++ b/src/components/ui/DataTable.tsx
@@ -7,7 +7,7 @@ import Dropdown from './Dropdown'
 import IconBtn from './IconBtn'
 
 /** Tailwind responsive breakpoints */
-export type TailwindBreakpoint = 'sm' | 'md' | 'lg'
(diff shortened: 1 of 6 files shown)


advplyr:
I removed the chapter ID column. That should always be incremental and so isn't useful. At some point early on it was different and had some use to it.

Thanks!

PR #228, opened 2026-08-25
Use Link for internal page navigation

### LINKED ISSUE #227: Playback stops when following links to other pages
**Steps to reproduce**

- While playing a book, navigate to a library item page.
- Click a series, author, or one of the detail links(genre, narrator, etc.)

**Expected:** the link navigates to the correct page and the book keeps playing.
**Actual:** the link navigates to the correct page, but the app reloads and playback stops.

The same issue happens on the library stats page (top genres, top authors, longest items, largest items) and on the narrators page.

hypnotoad08:
Fixes #227

Several internal links were plain hrefs, which trigger a full page load instead of client-side navigation. Because of this, the book that was playing stops.

Updated the item page (series, author, and the linked detail fields), the library stats page (top genres, top authors, longest items, largest items), and the narrator table to use Link.

The stats page links were also relative paths, which Link doesn't resolve the same way, so I made them absolute.

**Testing:**

- Played a book and confirmed it keeps playing when clicking links on the item, stats, and narrators pages
- Confirmed each link still lands on the right page with filters applied.


mikiher:
Thanks!

PR #234, opened 2026-08-25
Fix cover hover overlay after select/deselect

### LINKED ISSUE #225: Edit/Play cover buttons don't appear after select/deselect
The edit/play button on covers don't appear on hover after selecting/deslecting. In order to have them reapper we need to hover off the cover and then back.

### Expected

When using the multi-select then cancelling, hovering over the cover should show the edit/play buttons

### Actual

The edit/play buttons don't display after cancelling the multi-select until going out of focus/back to the cover.

https://github.com/user-attachments/assets/b3d0edd8-91f6-41ba-a401-0fa8579fa2bc

mikiher:
Fixes #225

- Restore play/edit overlay when leaving multi-select while the pointer is still on the cover
- Keep overlay visible when closing the more menu if the card is still hovered


advplyr:
works for me, thanks!

PR #239, opened 2026-08-26
Persist bookshelf cover size

### LINKED ISSUE #235: [BUG] - Cover Size setting set to 120 and resets when switching libraries
Hi

The Cover Size control is defaulting to 120 and if I set it to 100 it does not retain its selected value when switching between libraries.

On opening any library it shows as 

Image

### Steps to reproduce

Navigate to the react version of the server
login if needed
Use the Cover Size control (bottom-right corner) to change the cover size from the default value which seems to be set to 120
Image
Switch to a different library.
Cover size shows as 120
Return to the original library 
Cover size is back to 120

### Expected behaviour

The selected cover size should be preserved when switching libraries (either globally or per-library, depending on intended design). and should be 100 by default

### Actual behavior

The Cover Size setting resets to 120 whenever a different library is selected.

### Additional information
The issue occurs with the Cover Size control located in the bottom-right corner of the library view.
The control initially opens at 120 and reverts back to 120 after changing libraries.
Should default to 100

Tried on Edge and Firefox

hypnotoad08:
Fixes #235

The cover size widget held sizeIndex in local state, so the size reset when switching libraries or reloading.

Updated to store the size in a cookie after the discussion in https://github.com/advplyr/audiobookshelf/pull/5509.

The size is now available during the server render, so the covers paint at the saved size instead of the default. I moved CardSizeProvider into the (main) layout so it reads the size itself, instead of the widget pushing a multiplier in through an effect.

Desktop and mobile sizes are stored separately so adjusting one doesn't overwrite the other. The client records its viewport in a cookie so the server render knows which one to use, since the user agent can't tell a resized desktop window from a real desktop.

On a first visit, there is no viewport cookie yet, so the user agent is used as a fallback. That is correct for phones and desktops, but a desktop browser already narrowed to mobile width renders at 120 and then corrects to 100 after hydration. Every load after that reads the cookie and renders correctly from the first paint.

This branch had also stopped the bookshelf from clearing its loaded items when the page size changes, which 778eaece has since fixed upstream, so that part is gone. Keeping the items does expose a race the wipe was hiding. A fetch that was already in flight when the page size changes marks its page as loaded against the new boundaries, so some slots end up flagged as loaded while still empty and never get fetched. The remaining change ignores those late responses.

Testing

Changed the size, switched libraries, reloaded; size holds in both cases
Loaded the client on my phone and confirmed the mobile size persists separately from the desktop one, and renders at the saved size on first load
Confirmed the covers render at the saved size instead of the default on both desktop and mobile


mikiher:
Thanks for sending a fix, but I think I'd like to take a different approach.

While I was testing this, I noticed that when you load the homepage (from scratch), with a different saved size from the default 120, the shelves flash with the default size before re-rendering with the saved size. This is because the homepage is mostly SSRed, so the server rendered version can paint very quickly, and then it takes a 200-500 ms (on dev server) until the client is fully hydrated (with cover size loaded from local storage). This is different from the Vue client which didn't use SSR at all. I'd really like to avoid this flash, but I'd also like not to delay painting the page until full hydration.

This means localStorage is maybe not the best option for saving this. Instead, I'd like to save this on server user settings, this way, the value would be known at server render time, and we would be able to paint the homepage immediately (and not delay it until the page is hydrated).

This requires a small server change, but I think it's worth it, to keep the homepage loading swiftly and without flickers.

*Note: it's also possible to use client-side cookies to store that value, which doesn't require a server change. I think I like the user server settings options better, because it will open the door for persisting other user settings on the server, and using them for SSR, without polluting requests with many cookies.*

hypnotoad08:
No worries, I also like that approach better. I did notice the flash, but it appeared to be due to the js chunks loading via http/1.1 pushing hydration back. The cover images themselves also feel slower to populate then the vue client, however I have not tested the react client behind a ngnix proxy which would upgrade the connection to http/2.

mikiher:
I'm doing all my development and testing via reverse proxy. I've noticed that the homepage loads quite quickly (before this change), as well as the images taking a bit of time to load.

I am guessing the slow image loading on first load might be a separate issue, but it might also be related.

Are you willing to take on the server user setting solution?

hypnotoad08:
Yes I will give it a shot.

mikiher:
Thanks!

hypnotoad08:
@mikiher I updated the PR with the new approach and submitted a server PR for the API changes.

hypnotoad08:
Updated the pr to reflect the new api updates

hypnotoad08:
ok updated the design to be cookies

mikiher:
Thanks! (and thank you for your patience with us changing our minds about this)

PR #238, opened 2026-08-26
Add listened time alongside total duration on collection and playlist

AryaBait7:
Adds total listened time alongside the existing total duration shown on collection and playlist detail pages, addressing part of #189.

Both useCollectionBooks.ts and usePlaylistItems.ts already computed total duration client-side from data already loaded on the page. This adds a parallel calculation for listened time, using user.mediaProgress (already loaded app-wide) — summing each item's full duration if finished, or currentTime if still in progress.

No backend changes needed — both pages already load their full book/item list in one call, so this is purely a frontend calculation.

Tested locally:

Verified duration-only display still works correctly when nothing has been listened to.
Verified listened time updates correctly after playing a test book partway through, on both a collection and a playlist.

Collections Listened Time:

CollectionMins

Playlist Listened Time:

PlaylistMins

AudioBook Used for testing :

testbook3

Issue #189, opened 2026-08-12
feat: show total duration and listened progress on series page


mikiher:
Show total series duration and total listened duration on the series toolbar (alongside book count). Partially addresses advplyr/audiobookshelf#711.

**Original PR:** advplyr/audiobookshelf#5064  
**Notes:** May require server/API support (`getTotalDurationById`, progress aggregation). Should also be applied to playlists and collections (they already show total duration, but not listened/total).

@Vb2341 — would you like to port this to the React client?

mikiher:
@AryaBait would you like to take this?

If you do, I'd like you to verify the following things first:
1. Does it actually require new server API entrypoints? How many new server calls are required, if at all
2. Does `listened / total` on collection/playlist pages require new server calls, and how many are needed
3. Does this fit well on mobile toolbar for series/playlist/collection (assume a 360 pixel wide screen)?

AryaBait7:
Thanks for the detailed questions — I dug into the code, tested live on my local instance, and checked the precedent PR (advplyr/audiobookshelf#5064) and the original issue (advplyr/audiobookshelf#711) before answering:

1. New server API entrypoints? No new route needed for series — (advplyr/audiobookshelf#5064) shows the original implementation extended the existing endpoint (/api/libraries/:id/series/:seriesId, the same call the client already makes when opening a series) rather than adding a new one. It added a getTotalDurationById SQL sum on the Series model plus a loop over each book's progress for listened time. The current backend already has a very similar SQL pattern in seriesFilters.js (used today for sorting series by duration), so this isn't new infrastructure — it's extending a pattern that already exists. Net result: 0 new network calls from the client, but real server-side code is needed.

2. Listened/total on collection/playlist pages? Confirmed both in code and live: 0 new server calls. useCollectionBooks.ts / usePlaylistItems.ts already sum duration client-side today (I saw "2 Books (0m)" live on my test collection, screenshot attached) because those pages load every book in one call already. Listened time works the same way, since user.mediaProgress (with currentTime per item) is already loaded app-wide.

<img width="1900" height="955" alt="Image" src="https://github.com/user-attachments/assets/1ed2557d-217b-4ac2-9a64-980cd554f4af" />



[...]

PR #240, opened 2026-08-26
Add total duration and listened time to series page toolbar

AryaBait7:
Extends the pattern from #238  to the series detail page. Adds total duration and listened time next to the item count in the series toolbar (e.g. "Demo Series (3) (1m / 3m total)").

Backend: adds `totalDuration` and `progress.totalListened` to the existing `GET /libraries/:id/series/:seriesId` response in the paired `audiobookshelf` PR (advplyr/audiobookshelf#5502) — computed in-memory from the library items already fetched for that endpoint, so no new queries or network calls.

Frontend: reads the new fields in `SeriesClient.tsx` and renders them via the existing `itemCountSupplement` toolbar slot (same mechanism used for collections/playlists). Also fixes `Toolbar.tsx`, which was reading `itemCountSupplement` from context but never rendering it in the series-detail branch.

Tested locally against a Demo Series with 3 books, one with real listened progress — confirmed correct duration/listened math and that the toolbar updates live.

Part of #189.

SeriesPage:

SeriesPage


mikiher:
I'd like to get @advplyr 's opinion about #238 before I review this one.

PR #249, opened 2026-08-27
Serve the Next client from a configurable base path without a rebuild

### LINKED ISSUE #217: Support dynamic base path
in #112 @advplyr wrote:

> ... We discussed this briefly but it will probably not be as straightforward as I thought.
For context, we wanted to allow setting a custom base path like "https://yourserver.com/audiobookshelf" where "/audiobookshelf" is configurable.

> In Vue we settled with supporting a base path but forcing it to be "/audiobookshelf"

In our brief Discord discussion I wrote:

> ... it seems like it is not much easier in Next than it was in Nuxt. basePath in Next is **also a build time configuration**. Next seemlessly handles some of the cases (e.g. router.push/replace, Link, redirect), but in some of the cases (e.g. Image, fetch, new URL) you need to deal with it yourself (i.e. explicity add the base path to the URL you create).

So it looks like while Next has basePath support, it is a build time config and cannot be modified after build.
It does seem, though, that with SSR it is possible to create a dynamic solution.

For resolving this issue, I'd like to set a couple of requirements:
1. A dynamic basePath that doesn't require a build.
2. Callers of various URL/fetch functions do not need to know about it, or at the least use some standard URL transformation.

mikiher:
Fixes #217

Next inlines `basePath` into production chunks, so a simple `bastPath` deployment would have required a rebuild to change it.

This PR builds with a placeholder basePath, rewrites `.next` at server start from `RouterBasePath`, and prefixes the URLs Next does not prefix automatically (`fetch`, `next/image` `src`, Socket.IO, uploads, `window.location`). 

Root deployments keep `basePath` empty and behave as before.

This depends on [server PR #5507](https://github.com/advplyr/audiobookshelf/pull/5507) (restore the Express mount prefix when handing requests to Next).

## Testing
Run `pnpm build`

Run tests with 'npm run start-dev` (which loads a prebuilt next, proving that this doesn't require a rebuild)

Test with different values for `RouterBasePath` in `dev.js`, including empty string, and unset. 
Restart the server after each `RouterBasePath` change.
- Verify that when RouterBasePath is unset client is served under `/audiobookshelf` (similar to Vue)
- Verify that when `RouterBasePath == ''` client is served under `/`
- Verify that when `RouterBasePath == '/abs'` client is served under `/abs`

### DIFF
--- a/cypress/tests/lib/basePath.cy.ts
+++ b/cypress/tests/lib/basePath.cy.ts
@@ -1,4 +1,4 @@
-import { BASE_PATH_ATTRIBUTE, BASE_PATH_PLACEHOLDER, getClientBasePath, normalizeBasePath, replaceBasePathPlaceholder, withBasePath } from '@/lib/basePath'
+import { BASE_PATH_ATTRIBUTE, BASE_PATH_PLACEHOLDER, getBasePath, normalizeBasePath, replaceBasePathPlaceholder, withBasePath } from '@/lib/basePath'
 
 /** The browser reads the base path off the root element the server layout renders. */
 function serveFrom(basePath: string | null) {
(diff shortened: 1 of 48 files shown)


advplyr:
This works and makes sense. I still have to test OIDC though. I made one adjustment to throw an error in the case where the placeholder files aren't found and the base path needs to be updated.
Thanks!

PR #250, opened 2026-08-28
Move Upload to a library siderail page scoped to the selected library

### LINKED ISSUE #173: Suggestion: make Upload a library page

mikiher:
Fixes #173 

This removes the Upload button from the AppBar/UserAppBarNav and puts a library specific upload button on the library siderail.
The upload page moves under the library tree, and the only notabable change in it is that it doesn't contain the library selection dropdown anymore (since it is now a library specific page).


advplyr:
Adding that new button caused a regular desktop screen size to get a scrollbar in the siderail. I decreased the height of the buttons for that reason.

I think it looks worse cluttering up the siderail, but the only other option I could think of is to move the uploader into settings and bring back the library dropdown.
I'm sure a lot of people don't use the uploader at all, so we might get some push back. We'll see what people think.

PR #251, opened 2026-08-29
Make MediaSession respect chapter track setting

### LINKED ISSUE #191: feat(player): MediaSession respects chapter track setting
When "Use chapter track" is enabled, report chapter-relative duration/position to the Media Session API so OS lock screen and notification scrubbers span only the current chapter. Implements advplyr/audiobookshelf#5079.

**Original PR:** advplyr/audiobookshelf#5080  
**Notes:** Make sure this works in a similar way to other audiobook apps (including audiobookshelf-app)

@n8jadams — would you like to port this to the React client?

mikiher:
Fixes #191 - when "Use chapter track" is enabled, the Media Session API reports chapter-relative duration/position so OS lock screen and notification scrubbers span only the current chapter. 

### Notable changes from Vue
- Lock screen metadata in chapter-track mode uses Audible-style layout: book title primary, chapter name secondary, author in album (Vue uses book title as primary, author as secondary, no chapter title)
- Position updates subscribe to the progress store directly so playback ticks do not re-render `MediaPlayerContainer`

### Not yet implemented
- Share player chapter-track Media Session support
  - Currently Share player is not identical to Vue and doesn't show the Settings button that lets you control "Use chapter track"
    - **Should we add a Settings button on the Share player**
- Always showing chapter name on lock screen when chapter track is off
  - When "Use chapter track" is off, we show title and author on the lock screen. 
    - **Shouldn't we show title and chapter (if chapters exist) even when "Use chapter track" is off? Audible does it, and I think it makes more sense. There's not much info gain in showing the author**

### DIFF
--- a/src/app/(public)/share/[slug]/SharePlayer.tsx
+++ b/src/app/(public)/share/[slug]/SharePlayer.tsx
@@ -23,6 +23,7 @@ const PROGRESS_SYNC_INTERVAL = 30 // seconds
 
 export default function SharePlayer({ slug, startTime: startTimeParam }: SharePlayerProps) {
   const t = useTypeSafeTranslations()
+  const unknownLabel = t('LabelUnknown')
   // Share data
   const [shareData, setShareData] = useState<MediaItemShareResponse | null>(null)
   const [fetchError, setFetchError] = useState<string | null>(null)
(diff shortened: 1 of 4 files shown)


advplyr:
I think it's probably better to put the chapter title as title in all cases if it has a chapter title. Then we can keep artist to be more accurately set as the author. Album can be used for the book title.
Another option is to do this like we do the minimized mobile app audio player where we append the chapter title to the book title. `Book Title: Chapter title`. I think a lot of displays should scroll longer text but I would have to test that.

mikiher:
I like the second option better. The probem with the first is that Album is never displayed anywhere (at least on Android), So you don't see the book title, which seems weird for audiobooks (no other audiobook player I saw just displays chapter title without book name).

mikiher:
OK, I implemented the second option (including the share player). PTAL again.

advplyr:
The media session API can be used by any device, so my hesitation with using artist for chapter title is that a device could  display like `Aritist: <insert media session artist>`. I'm pretty sure I've seen car dashboards display it like that.

advplyr:
Thanks!

advplyr:
I tested with a pixel 9a and the long title doesn't scroll which is unfortunate. On iphone 15 it scrolls the title

mikiher:
It scrolls on my Samsung S24, so I guess it will on most modern Samsung phones.

PR #256, opened 2026-08-29
Underline inline links when hover is unavailable

### LINKED ISSUE #219: Make link styiling consistent
Currently many links in the app show an underline when hovered, but there's no hover on mobile, so there's no indication that this is a link until you click on it.

In some places this is handled (usually by making the underline permanent on mobile), but the treatment is not consistent.

We should make link styling consistent across the app (both a and Link elements)

mikiher:
fixes #219

## Notable changes
- Add a `link-underline` utility
   - always underlines when hover is unavailable
   - underlines on hover when it is (`@media (hover: hover)`).
- Apply it to inline `<Link>` / `<a>` text (and similar clickable table cells)

Card/row/menu surfaces, icon-only links, and already-permanent underlines (e.g. settings footer) are unchanged.

## Testing
- Mobile: author, series, stats, player metadata, and similar inline links are underlined at rest
- Desktop with a mouse: those links have no underline at rest and underline on hover
- Collection/playlist list view on mobile: series, author, and episode subtitle are tappable links
- Cover cards, search results, notification task rows, and help icons still have no underline

### DIFF
--- a/src/app/(main)/library/[library]/(book)/authors/[author]/AuthorClient.tsx
+++ b/src/app/(main)/library/[library]/(book)/authors/[author]/AuthorClient.tsx
@@ -107,11 +107,11 @@ export default function AuthorClient({ author: authorProp }: AuthorClientProps) 
         <div className="mt-8e -ms-2e">
           <ItemSlider
             title={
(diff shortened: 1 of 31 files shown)


advplyr:
Works great, thanks!

PR #262, opened 2026-08-30
Distinguish dropdown selected state from focus and hover

### LINKED ISSUE #246: Embed Metadata is highlighted always on this menu.
No matter which tool is selected, "Embed Metadata" is highlighted. 
Image

mikiher:
Fixes #246

Opening a dropdown by click no longer looks like the first item is chosen. Keyboard focus, hover, and the actual selected value are separate, including nested filter items.

### Notable changes

- Click-open leaves items unfocused; 
  - Arrow Down / Enter / Space focus the first item
  - Arrow Up the focuses the last
- Selected value uses the theme selected color
  - a submenu choice highlights both the parent and the subitem (see library filter dropdown)
- Theme colors:
  - `dropdown-item-focused` for the focus fill (applied to bg)
  - `dropdown-item-selected` for the current-value color (applied to text, darker on light theme)

### DIFF
--- a/cypress/tests/components/app/UserAppBarNav.cy.tsx
+++ b/cypress/tests/components/app/UserAppBarNav.cy.tsx
@@ -1,8 +1,8 @@
 import UserAppBarNav from '@/app/(main)/UserAppBarNav'
 import { UserContext, type UserContextType } from '@/contexts/UserContext'
 import { User } from '@/types/api'
-import { AppRouterContext } from 'next/dist/shared/lib/app-router-context.shared-runtime'
 import type { AppRouterInstance } from 'next/dist/shared/lib/app-router-context.shared-runtime'
+import { AppRouterContext } from 'next/dist/shared/lib/app-router-context.shared-runtime'
 import * as navigation from 'next/navigation'
 import { ReactNode } from 'react'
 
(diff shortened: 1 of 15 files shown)


advplyr:
The checkbox is no longer right-aligned in the multi select dropdown.

image

Using the arrow keys no longer works for the multi select dropdown. Well it does appear to make selections but they aren't visible/highlighted. Same with regular dropdowns.

The intention is that if you click to open the dropdown you can still arrow down to select, right? Or is the intention to completely remove keyboard navigation if you have clicked to open? I think preferably the arrow keys can always be used.

mikiher:
Yes that's the intention. Sorry I broke the multi select, will fix

mikiher:
OK, I fixed a couple of issues:

1. I removed the checkboxes altogether (standardizing on text-dropdown-item-selected for selected, in both regular and multi-select dropdowns). If you want the checkboxes back, I can add them back, but wanted to make selected-style consistent.
2. I accidentally added code that moved keyboard focus when hovering over an item. I removed that code.

I couldn't, however, repro your complaints before or after my changes. Anyway PTAL again. Make sure you reload the server before your checks - I'm not sure if changes in globals.css are properly picked up by HMR.

advplyr:
I agree, the checkboxes aren't necessary. This is working well for me now, thanks!

Issue #233, opened 2026-08-25
Dirty tracking and confirm unsaved changed in Details and Match sections


mikiher:
As discussed in a [pr 210 comment](https://github.com/audiobookshelf/audiobookshelf-client-react/pull/210#issuecomment-5408125451), We should implement dirty tracking and confirm unsaved changes in the Details and Match sections in the Edit Modal.

PR #266, opened 2026-09-03
Add Embed button to the metadata edit modal footer

### LINKED ISSUE #247: Embed is not in the Edit Menu.
I assume this is intended but it makes my edit process difficult. Normally I would like to go from Match to embed in two clicks in the old ui. Now I have to get out of the match menu, click the three dots, look for "tools", click embed, and then "start Embed".  This is much slower and more cumbersome. I would prefer the old page which had tools as a tab in the edit window.

Image

If we can't have an embed page built into the edit side bar another compromise would be to allow edits to get auto written to the files so I don't have to go through the repetitive process that requires a lot more clicks than it did previously.

mikiher:
Closes #247.

Puts Embed, Preview & Embed, and Make M4B in the shared library-item edit footer so admins do not have to leave the modal (or hunt for a Tools tab) after matching or editing metadata. 

The single button design with stored selected function is drawn from Github Close Issue button.

*Note: please review functionality, not code. This not ready for merging yet.*

*Note: mobile is not fully considered yet. Check only on desktop for now.*

### Notable changes
- **Shared footer** across Details, Cover, Chapters, and Match. 
  - Section actions (Save, Discard, Match Submit) 
  - Embed control stays on the start side.
- **Split control** for admin users on books with audio tracks
  - primary runs the last chosen action
  - chevron opens the three options. 
  - The selected option is persisted in `localStorage`.
- The button options:
  - **Embed** (was Quick Embed) writes metadata in place after the backup confirmation
    - shows embed progress
  - **Preview & Embed** (was Embed Metadata "Open Manager" button) closes the modal and opens the embed page.
  - **Make M4B** (was Make M$B "Open Manager" button) closes the modal and open the m4b page.
- **Dropdown** 
  - Added support for:
    - end-aligned menus
    - stacked option descriptions
    - wider min width on small screens so subtext does not truncate.


advplyr:
I don't think that UI works well in this case. Having some actions open to a page and the other show a confirmation makes it feel off. Githubs usage of that UI works because all the options are different ways to close the pull request.

It *could* work with the Save & Close button. Having an option to use "Save & Quick Embed" or "Save & Close". I'm not sure I like that either though.

My first thought is trying a context menu type button in the bottom left of the side rail that has action items for quick embed and go to tools page. I don't have a great solution in mind though.

advplyr:
> My first thought is trying a context menu type button in the bottom left of the side rail that has action items for quick embed and go to tools page

I played around with this for a few minutes and I don't think that will work.

mikiher:
What about only a quick embed button, without the other two go-to-page actions?

We can think where to put those later but at least there will be a quick embed option?

advplyr:
Yeah I suppose we could do that. Previously we had a quick match there

mikiher:
OK, switched to Embed button only (force-pushed the modified commit).

I stashed the Split Button component so maybe we can use it in the future. for something else (I removed it from the submitted branch).

PTAL again, and let me know if this works for you.

advplyr:
I think we should only have that button on the details tab. We don't need a shared bottom toolbar for that modal. The modal was designed so that each page is independent from the others.
We could have matching go to the details tab after saving. Currently there is no benefit that I can see for a match to return to the search page. So if saving a match goes to the details tab, then the best place for the button is on the details tab.

We can also use "Quick Embed" since that is more accurate

mikiher:
> I think we should only have that button on the details tab. We don't need a shared bottom toolbar for that modal. The modal was designed so that each page is independent from the others. We could have matching go to the details tab after saving. Currently there is no benefit that I can see for a match to return to the search page. So if saving a match goes to the details tab, then the best place for the button is on the details tab.

I understand why you want matching to end up in the the details tab (In an ideal world, Match should be inside the details tab, and not be a separate tab).

However:

1. I do not underand why the chapters and cover tabs should not have an Emebd button. they are editing different aspects of the book's metadata. Why should I not have Embed at my disposal after editing chapters or cover? how are they different from details?
2. I think there's a benefit in having a shared footer. Let's put aside Match for a moment, as it is somewhat different - shouldn't Details, Cover, and Chapters all eventually have Save and Discard? It's a much cleaner model - you make changes, then commit or discard. 

> 
> We can also use "Quick Embed" since that is more accurate

I think the "quick" prefix is also somewhat inaccurate. You still have to go through a confirmation dialog with an option that needs your attention. 

I really think that there should just be an Embed button with a confirmation dialog, and no Embed page. From a pure functional perspective, the Embed page doesn't add anything on top of that - it's full of info you can't do anything about, and the only controls it offers are already available in the confirmation dialog. 

If you want to explain what Embed exactly does (which is useful for nitpickers to open new bugs against us), you can add an info link. There's really no gain in maintaining a page for all this fluff.

advplyr:
This is getting into what we discussed originally when trying to figure out how to keep the metadata editing inside a modal. We moved away from tabs because each one has to be independent in order to avoid redesigning the entire experience. If we were to redesign metadata editing so that they all have a save and discard button, then we could just use the original tabbed modal.

The main reason I don't like the embed button on all of the tabs is because it looks bad. It looks particularly bad on the Cover and Match tabs because there is nothing else down there. So it was purely aesthetics.

I agree we could get rid of the embed page. The benefit of that page is we show what we are about to embed. I'm not sure how helpful that is to users. The confirmation modal for embed could be larger and include the meta tags.

mikiher:
> This is getting into what we discussed originally when trying to figure out how to keep the metadata editing inside a modal. We moved away from tabs because each one has to be independent in order to avoid redesigning the entire experience. If we were to redesign metadata editing so that they all have a save and discard button, then we could just use the original tabbed modal.

To me it looks like we've already brought back tabs to desktop, in a nicer package (and it is nicer).

There's already a bug (#233) around save guards in the Details and Match pages, and I was about to open a bug about putting Save and Discasrd on the Cover page. I think it would work nicely. I think unifying the edit flow in the Details, Cover, and Chapters tabs (Save, Discard, Dirty coloring, Save guards) is a worthy cause and not very hard to implement.

> 
> The main reason I don't like the embed button on all of the tabs is because it looks bad. It looks particularly bad on the Cover and Match tabs because there is nothing else down there. So it was purely aesthetics.

As I said, Match is different - you said you want post-submit to switch to Details, you got it - it doesn't need an Embed 
button. 

For Chapters and Cover, the plan is to eventually have Save and Discard for both, so there will be something down there, and the Embed button will not be lonely.

> I agree we could get rid of the embed page. The benefit of that page is we show what we are about to embed. I'm not sure how helpful that is to users. The confirmation modal for embed could be larger and include the meta tags.

My opinion is that since it's not actionable (can't edit), it's not worth showing. You already know *what* you're embedding (all the embeddable metadata), and if you're interested in which metadata field is mapped to which metadata tag, you should read the documentation.

If we were to allow editing the metadata before embedding, it would have been a different story, but right now it's just read-only, and you can't do anything about it other than abort.

advplyr:
Can you clarify if you are thinking that this would function like the tabbed modal functions?

mikiher:
> Can you clarify if you are thinking that this would function like the tabbed modal functions?

Sorry, I'm not sure I understand. In what sense? I think we might be talking past each other. Let's discuss over discord.

mikiher:
OK, after discussing, I understood @advplyr prefers keeping existing design where each section keeps its own state, and there's no sharing between sections. I'm going to close this PR now since it has veered too far from my initial plans. I'll get back to it after I implement changes to make the Details, Cover and Chapters sections be more similar to one another.

PR #268, opened 2026-09-04
Confirm unsaved changes when closing Details tab in edit modal

SharonZhu0122:
Related to #233

The Chapters tab already warns before discarding unsaved edits when the modal is closed or the section is switched. The Details tab had no such guard — editing the title, authors, genres, explicit/abridged flags, etc. and then closing the modal silently discarded the changes with no warning.

This adds the same requestLeave + confirm-dialog pattern already used by ChaptersEditModalBody to LibraryItemEditModalContent, and updates the parent modal to check whichever section (details or chapters) is currently mounted before allowing navigation away.

The Match tab still needs the same treatment — happy to follow up with that in a separate PR if useful.

**Testing:**
- Edited the title, authors, genres, and explicit/abridged checkboxes in the Details tab, then closed without saving — confirm dialog now appears (previously it did not)
- Confirmed Save / Discard / Cancel in the dialog all behave correctly
- Confirmed switching tabs (e.g. to Chapters) while Details has unsaved edits also triggers the confirm dialog
- Confirmed editing only whitespace (e.g. trailing spaces in the title) does not trigger a false-positive warning, since those fields are trimmed before comparison
- Confirmed the Chapters tab's existing unsaved-changes behavior is unaffected

### DIFF
--- a/src/components/modals/ChaptersEditModalBody.tsx
+++ b/src/components/modals/ChaptersEditModalBody.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useLibraryItemModal } from '@/components/modals/LibraryItemModal'
+import { useLibraryItemModal, type UnsavedChangesLeaveHandle } from '@/components/modals/LibraryItemModal'
 import Btn from '@/components/ui/Btn'
 import Checkbox from '@/components/ui/Checkbox'
 import HelpTooltipIcon from '@/components/ui/HelpTooltipIcon'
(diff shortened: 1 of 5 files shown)


mikiher:
Thanks!

This works well, but as you said it doesn't completely fix #233 
- Match section needs the same confirm dialog
- Dirty tracking on Details Edit is slightly buggy (return to initial state doesn't clear dirty state)
- I'd like to implement dirty coloring similar to Chapters Edit

PR #270, opened 2026-09-05
fix(chapters): keep first chapter start fixed at zero

Neon1201:
## Brief summary

Keeps the first chapter start fixed at `00:00:00` in the chapter editor.

## Which issue is fixed?

Fixes an inconsistency where Chapter #1 could be edited away from `00:00:00` even though validation requires the first chapter to start at zero.

## In-depth Description

The chapter editor previously allowed users to modify the start timestamp of Chapter #1.

This created an invalid editor state because the first chapter is required to start at `00:00:00`, and attempting to save a non-zero first chapter start resulted in a validation error.

This change:

- Keeps the first chapter start normalized to `00:00:00`
- Disables the first chapter timestamp controls in the UI
- Prevents shared editor update functions from moving the first chapter away from zero
- Preserves all later chapter start times
- Keeps the existing start-only chapter boundary model unchanged

## How have you tested this?

- Relevant Cypress tests: 9 passing
- `pnpm typecheck`
- ESLint on changed files
- `prettier --check` on changed files

Regression coverage includes:

- non-zero first chapter start is normalized to `00:00:00`
- later chapter start times remain unchanged
- first chapter timestamp controls are disabled
- later chapter timestamps remain editable
- chapter end calculation remains correct

### DIFF
--- a/cypress/tests/components/widgets/chapters-edit/ChapterEditTableRow.cy.tsx
+++ b/cypress/tests/components/widgets/chapters-edit/ChapterEditTableRow.cy.tsx
@@ -1,0 +1,75 @@
+import ChapterEditTableRow from '@/components/widgets/chapters-edit/ChapterEditTableRow'
+import type { EditableChapter } from '@/lib/chapters/chapterEditorUtils'
+import type { ComponentProps } from 'react'
+
+const mediaDuration = 3600
+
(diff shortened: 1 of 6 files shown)


mikiher:
Thanks!

PR #271, opened 2026-09-05
fix(chapters): detect one-second timestamp edits

### LINKED ISSUE #264: [Bug] Save button doesn't show up in new chapter editor when changing few seconds up/down
https://github.com/user-attachments/assets/7a98c898-1757-47c7-9cb0-7730e66c1131

As you can see in the video,  I have to edit some chapter's title (to make save button show up) to save my timestamp fixes.

Neon1201:
## Brief summary

Fixes chapter editor dirty detection so real one-second timestamp edits are recognized as unsaved changes.

## Which issue is fixed?

Fixes #264

## In-depth Description

The chapter editor previously used the same ±1 second timestamp tolerance for both imported chapter matching and user-edit dirty detection.

As a result, changing a chapter start time by exactly one second was treated as unchanged. The Save button remained unavailable until another field, such as the chapter title, was modified.

This change separates the two behaviours:

- User edits now use exact whole-second comparison after timestamp normalization.
- The existing ±1 second tolerance is preserved for Audible/import matching.
- Dirty row highlighting now matches the Save state.
- The existing chapter matching window and start/end model are unchanged.

## How have you tested this?

- Relevant chapter Cypress tests: 22 passing
- ESLint on changed files: passing
- Prettier check on changed files: passing
- Cypress TypeScript check: passing

Regression coverage includes:

- `100 -> 101` is dirty
- `100 -> 99` is dirty
- `90.4 vs 90` remains clean after normalization
- row dirty highlighting agrees with Save state
- Audible/import matching still preserves ±1 second timestamp tolerance

`pnpm typecheck` currently fails on master due to an unrelated existing `GlobalSearchInput.tsx` / `useEffectEvent` error.

### DIFF
--- a/cypress/tests/lib/chapterEditorUtils.cy.ts
+++ b/cypress/tests/lib/chapterEditorUtils.cy.ts
@@ -1,7 +1,9 @@
 import {
+  applyMatchedClientKeys,
   buildBulkChapters,
   buildChapterDirtyBaseline,
   buildIdenticalChapters,
+  computeHasChanges,
   detectBulkChapterPattern,
   getChapterDirtyFields,
   initChapters,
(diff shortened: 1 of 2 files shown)


mikiher:
Thanks!

PR #272, opened 2026-09-05
fix: dismiss topmost modal on browser back

### LINKED ISSUE #269: Browser Back should close open modals instead of navigating away
When a modal is open, browser/system Back should dismiss it and stay on the current page, the same as using the modal Close action. Today Back navigates (or collapses a layer under the modal) and can leave the dialog open.

This should apply to ordinary overlay dialogs app-wide.

### Steps to reproduce

1. Open any modal
2. Press the browser Back button (or the Android/system Back gesture).

### Expected behavior

The modal closes. The current page stays as it was. A second Back can then navigate.
If more than one modal is open, first back should close the top modal, second back should close the one below it, etc.

### Actual behavior

Back leaves the page.

*Note: the expected behavior should be identical to clicking Close or pressing Escape*

blue219:
## Brief summary

Browser Back now dismisses the topmost ordinary modal without navigating away from the current page. Nested dialogs close one at a time; Back navigates normally once all dialogs are closed.

## Which issue is fixed?

Fixes #269.

## In-depth Description

- Integrate shared history handling into `Modal`, adding a same-URL history entry per open layer while preserving existing history state.
- Invoke the existing close callback so business confirmation and cleanup behavior remain intact. Persistent or processing dialogs block Back, consistent with their Escape behavior.
- Reconcile history after button, Escape, backdrop, save, and programmatic dismissal. Coordinate asynchronous history cleanup, Strict Mode effects, and stale entries encountered after Forward or reload.
- Give modal dismissal priority over the unsaved-page navigation guard, and defer guard history changes until modal entries have been removed. Real navigation is not undone by modal unmounting.

## How have you tested this?

- Added a regression test using real browser history and confirmed it failed before the fix because the modal remained open after Back.
- All 16 Cypress component tests pass, covering nested dialogs and preserved drafts, repeated dismissal through different close actions, persistent/processing dialogs, confirmation callbacks, declined dismissal, queued Back operations, Forward, navigation/unmount cleanup, and unsaved-guard setup/removal.
- TypeScript checks pass for the application and Cypress projects. Targeted ESLint and `git diff --check` pass.
- Manually verified the local Next.js application in Chrome: Back closes ordinary and confirmation dialogs without changing the URL or scroll position; persistent dialogs stay open; reloading does not reopen dialogs; subsequent Back returns to the preceding page.
- Android system Back gestures have not been tested on a physical device.

Commands used:

```sh
node node_modules/cypress/bin/cypress run --component --browser chrome --spec cypress/tests/components/modals/ModalHistory.cy.tsx
node node_modules/typescript/bin/tsc --noEmit --incremental false
node node_modules/typescript/bin/tsc --noEmit --incremental false -p cypress/tsconfig.json
node node_modules/eslint/bin/eslint.js src/hooks/useModalHistory.ts src/hooks/useUnsavedNavigationGuard.ts src/components/modals/Modal.tsx cypress/tests/components/modals/ModalHistory.cy.tsx
git diff --check
```

### DIFF
--- a/src/hooks/useDetailsEdit.ts
+++ b/src/hooks/useDetailsEdit.ts
@@ -83,6 +83,23 @@ export interface DetailsEditRef<TDetails> {
   mapBatchDetails: (batchDetails: Partial<TDetails & { tags: string[] }>, mapType?: 'overwrite' | 'append') => void
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function applyTrimFieldsToDetails<TDetails extends Record<string, any>>(details: TDetails, trimFields?: ReadonlyArray<keyof TDetails>): TDetails {
+  if (!trimFields?.length) return details
+
+  let next: TDetails | null = null
+  for (const key of trimFields) {
+    const value = details[key]
+    if (typeof value !== 'string') continue
+    const trimmed = value.trim()
+    if (trimmed === value) continue
+    if (!next) next = { ...details }
+    next[key] = trimmed as TDetails[typeof key]
+  }
+
+  return next ?? details
+}
+
 interface UseDetailsEditOptions<TDetails> {
   metadata: TDetails
   tags: string[]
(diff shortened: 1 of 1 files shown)


mikiher:
Why did you close this?

blue219:
> Why did you close this?

I don't know what happened either, I must have been drunk, I submitted a new pr instead

PR #273, opened 2026-09-05
Mini-player and Fullscreen player redesign

mikiher:
This is a **major player redesign**, focusing mostly on mobile (but including several desktop polish items). The old split mobile/desktop layouts (`PlayerMobileLayout`, `PlayerDetailsPanel`, etc.) are replaced by a single **Player Shell** that morphs between mini and fullscreen.

The new fullscreen view is designed to be super-responsive and render well on every mobile size, portrait and landscape.

*Note: please review functionality, not code. This not ready for merge yet*

### Mini player (bottom bar)

- **New layout** with partial functionality
  - small cover on the left, title/author, compact transport (jump back + play/pause only), and track bar.
  - no secondary toolbar
  - no prev/next chapter
- **Slide-up animation** when playback starts; the page inset grows with it.
- **Slide-down animation** when closing (swipe down or close)
  - audio **stops immediately** on close, then the player animates away.
- **Page layout** now uses the player’s real measured height instead of a fixed ~165px guess, so content above the bar fits more accurately.
- **Long titles and chapter names** can scroll horizontally (marquee) when they don’t fit.
  - similar to mobile app
- **Close button** on mobile mini bar is hidden on mobile
  - Reasons:
    - save vertical space
    - not provide a Close touch target close to the play button
    - same as mobile app
  - Can be closed via swipe down. 
  - On desktop (lg+), close stays visible in the bar

### Fullscreen player (mobile)

- Replaces the old “expanded details” panel with a **true fullscreen overlay** (cover, metadata, controls, secondary toolbar).
- **Tap cover** or **swipe up** on the mini player to expand.
- **Swipe down** (or click on checron) on fullscreen to collapse back to mini.
- **Animated cover morph** when expanding/collapsing (view transition).
- **Collapse** button (chevron) in the top corner.
- **Secondary toolbar** (volume, playback speed, sleep timer, etc.) is available in fullscreen
- In **chapter-track mode**
  - fullscreen can show **two progress bars** (chapter + whole book)
  - chapter label is shown above the track bar (more space) or below (between the two seek numbers) depending on available vertical space.
- **Portrait**
  - Bottom block always shown fully
  - Cover size may shrink according to available height.
- **Landscape**: layout adapts according to available vertical space. c and d are not likely to happen on most modren phones.
  - a. single progress bar instead of two
  - b. move chapter label below the track bar
  - c. hide author and book duration line
  - d. hide secondary toolbar
  - Cover size may shrink according to available height.

### Gestures & seeking

- Vertical swipes work **over the whole mini player**
  - including the track bar (vertical gesture wins over seek).
- Same on fullscreen: swipe down won’t accidentally seek.
- Track bar supports **drag-to-seek** (pointer), not just click.
- Track bar is a proper **keyboard-accessible slider** (arrow keys, etc.).

### Navigation & keyboard

- **Browser/System Back**: 
  - if a player modal or volume/speed popover is open, Back closes that first
  - if no modal is open, collapses fullscreen
- **Escape** (when player hotkeys are active)
  - Same behavior as Back
- **Title/author links** collapse fullscreen before navigating away.
- While fullscreen, **rest of the app is inert** (can’t tab/click behind the player). 
  - E-reader overlay is also inert if open at the same time.
- Fullscreen player sits **above** the e-reader (higher z-index).

### Desktop (lg+)

- Still shows a **mini player**, but rebuilt with the new shell styling.
- Mini player shows
  - **full transport** (all chapter/skip buttons) 
  - **Close** button
  - **full secondary toolbar**
- Can expand to fullscreen (cover tap / mouse swipe)

### Other polish

- Cover art uses the **library’s cover aspect ratio**
- `prefers-reduced-motion`: enter/close animations are skipped.


advplyr:
This is definitely the right direction.



> * small cover on the left, title/author, compact transport (jump back + play/pause only), and track bar.

I think we should also include the jump forward. I know it isn't as useful but it looks odd without it. I like having the symmetry.

A few times while testing on mobile I tried tapping the mini player in the middle to expand. It may be because I'm so familiar with the mobile app or mini-players in other apps. We may want to make a tap anywhere not on a link or button expand on mobile.



> * **Slide-up animation** when playback starts; the page inset grows with it.
> * **Slide-down animation** when closing (swipe down or close)
>   
>   * audio **stops immediately** on close, then the player animates away.
> * **Page layout** now uses the player’s real measured height instead of a fixed ~165px guess, so content above the bar fits more accurately.

I think it would be better to not calculate the height of the player with js like this. The slide up/down on opening and closing adds a lot of complexity. If we don't have the slide up/down animation then we should be able to put the static mini player heights (for desktop & mobile) in a CSS variable.

> * Track bar supports **drag-to-seek** (pointer), not just click.

While testing this I noticed the seek was being called continuously while dragging. I think this is a good feature, but we should only call seek once on drag end.

mikiher:
> I think we should also include the jump forward. I know it isn't as useful but it looks odd without it. I like having the symmetry.

Fixed.

> A few times while testing on mobile I tried tapping the mini player in the middle to expand. It may be because I'm so familiar with the mobile app or mini-players in other apps. We may want to make a tap anywhere not on a link or button expand on mobile.

Fixed.

> I think it would be better to not calculate the height of the player with js like this. The slide up/down on opening and closing adds a lot of complexity. If we don't have the slide up/down animation then we should be able to put the static mini player heights (for desktop & mobile) in a CSS variable.

Fixed. removed slide-in/out animation and JS height calculation.

> While testing this I noticed the seek was being called continuously while dragging. I think this is a good feature, but we should only call seek once on drag end.

Fixed.

advplyr:
I found a few issues:

- The author name is duplicated

image

- When opening a player modal, like the player queue, then closing it by clicking outside the modal, the player gets expanded

- The player queue is always using non-square book cover aspect ratio

image

mikiher:
> * The author name is duplicated

Sorry, this was a side-effect of fixing the author-names marquee loop. Fixed.

> * When opening a player modal, like the player queue, then closing it by clicking outside the modal, the player gets expanded

Fixed.

> * The player queue is always using non-square book cover aspect ratio

Fixed.

advplyr:
I wasn't able to find any other bugs. 

For the UI I think we should increase the gap between the chapter track and the full track. The timestamps that are associated with the chapter track are actually closer to the full track.

image

I also noticed on the desktop player that the chapter track is 12px higher than the full track. You can notice it when toggling on the chapter track.

For the desktop player when minimized I think we can decrease the height. There is extra whitespace under the cover. We could also increase the width of `.player-title-author`

image


I experimented with 9rem (144px) player height on desktop, and increasing `.player-title-author` to 28% width on < 1280px and 32% width >= 1280px.

image

mikiher:
> I wasn't able to find any other bugs.
> 
> For the UI I think we should increase the gap between the chapter track and the full track. The timestamps that are associated with the chapter track are actually closer to the full track.

Fixed.

> I also noticed on the desktop player that the chapter track is 12px higher than the full track. You can notice it when toggling on the chapter track.

Fixed.

> For the desktop player when minimized I think we can decrease the height. There is extra whitespace under the cover. 

Fixed.

> We could also increase the width of `.player-title-author`
> I experimented with 9rem (144px) player height on desktop, and increasing `.player-title-author` to 28% width on < 1280px and 32% width >= 1280px.

I actually made it take all the space between the cover and transport (with padding).
Desktop Mini player height is now 136px (I was able to save a few more pixels by making the cover height match the title+author+duration height).

If this looks OK, give me a thumb up and I'll start working on making the code presentble. For some reason the AI chose to put stuff in `player-shell.css` module. Do you want it like this, or do you prefer standard tailwind classes like most of the code?

advplyr:
> > I also noticed on the desktop player that the chapter track is 12px higher than the full track. You can notice it when toggling on the chapter track.
> 
> Fixed.

This one is still an issue. It is easiest to see if you open the player settings and toggle the chapter track back and forth. You'll see the chapter track jump up 12px.

image


This is minor but I think the gap on these action buttons would look better as a fixed 8px or max 12px gap. The 16px just looks a bit off to me. I also recently went through and updated all our form input spacing to 12px and 8px for condensed.

image


> If this looks OK, give me a thumb up and I'll start working on making the code presentble. For some reason the AI chose to put stuff in `player-shell.css` module. Do you want it like this, or do you prefer standard tailwind classes like most of the code?

I don't think we need to try to get every single thing in a tailwind class, but it's better to use them where we can. Then leave the custom CSS for where it makes sense, like animations or CSS vars.

This functionality is great though. Big improvement over the current player.

mikiher:
> > > I also noticed on the desktop player that the chapter track is 12px higher than the full track. You can notice it when toggling on the chapter track.
> > 
> > 
> > Fixed.
> 
> This one is still an issue. It is easiest to see if you open the player settings and toggle the chapter track back and forth. You'll see the chapter track jump up 12px.

Sorry, I thought you meant on the fullscreen player (where it also displayed track bars with different heights).
It's now also fixed in the mini player. I also fixed the gaps between the cover and track bar so it matches cover left inset gap (16 px on desktop, 8 px on mobile). The mini player height is now 110px on mobile and 133px on desktop

> This is minor but I think the gap on these action buttons would look better as a fixed 8px or max 12px gap. The 16px just looks a bit off to me. I also recently went through and updated all our form input spacing to 12px and 8px for condensed.

I fixed the gap to 8px on desktop and 4px on mobile (action buttons are wider and higher on mobile, 44px, to make tap targtets big enough).

> > If this looks OK, give me a thumb up and I'll start working on making the code presentble. For some reason the AI chose to put stuff in `player-shell.css` module. Do you want it like this, or do you prefer standard tailwind classes like most of the code?
> 
> I don't think we need to try to get every single thing in a tailwind class, but it's better to use them where we can. Then leave the custom CSS for where it makes sense, like animations or CSS vars.

I already see some subtle conflicts in the player code where we have a mixture of both. I'll make sure to use mostly tailwind to clean this up.

> This functionality is great though. Big improvement over the current player.

Thanks!

PR #274, opened 2026-09-07
docs: reference server's npm

slagiewka:
Server is still npm based, referencing pnpm is likely to cause confusion.

Personally I'd advise users to just run `node --run ...` everywhere, which is package manager independent, but it requires node 22+, which the server doesn't do right now (uses v20 in Dockerfile etc.).


mikiher:
Thanks! (we will likely switch to a newer version of Node soon)

PR #275, opened 2026-09-08
Bring back +/- buttons for chapter timestamps

### LINKED ISSUE #265: [Feature] Bring back +/- button in new chapter editor
The new chapter editor has removed the +/- buttons that were present in the old client beside the chapter timestamps. They were very useful for making quick timestamp adjustments using only the mouse, without requiring a keyboard.

For example, say you have an old rip of an audiobook and you use Audible metadata matching in ABS. You find that the audiobook you have is 24 seconds shorter or longer than the matched version. However, it’s rarely the case that every chapter is exactly 24 seconds ahead or behind. In reality, some chapters might be 2 seconds ahead, some 8 seconds behind, some 10 seconds ahead, and so on.

I used to be able to easily fix these differences by playing each chapter and adjusting the timestamp a few seconds up or down by clicking those buttons until the timing was right. However, this has now become tedious because I have to type the number of seconds multiple times using the keyboard to find the right spot for every chapter.

In my opinion, this is a regression from the functionality that was available before.

lukawcf:
Closes #265. Adds small +/- buttons next to each chapter's start time so you can nudge it by a second without typing. They're disabled once you hit the start or end of the track.

### DIFF
--- a/src/components/ui/DurationPicker.tsx
+++ b/src/components/ui/DurationPicker.tsx
@@ -16,6 +16,8 @@ export interface DurationPickerProps {
   borderless?: boolean
   size?: 'small' | 'medium' | 'large'
   className?: string
+  /** Extra classes on the bordered InputWrapper (e.g. glued input groups). */
+  wrapperClassName?: string
   /** Accessible name for the time group. Defaults to LabelDuration. Ignored when ariaLabelledBy is set. */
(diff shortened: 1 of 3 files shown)


mikiher:
I've added some style and layout changes. Thanks!

PR #276, opened 2026-09-09
Confirm unsaved changes when leaving the Match tab in the edit modal

lukawcf:
Related to #233. Follows up on #268, which added the unsaved-changes guard for the Details tab; this does the same for the Match tab.

In the Match section a selected match is only applied on Submit, so switching sections, going back to the hub, navigating prev/next, or closing the modal with a match selected silently dropped it. Match now exposes the same requestLeave handle as Details and Chapters, and the sectioned edit modal checks it alongside the other two. The confirm dialog offers Discard / Cancel only, since applying a match the user has not finished reviewing did not seem like a safe default.

## Test plan
- [ ] eslint, prettier, tsc --noEmit and find-hardcoded-strings all pass
- [ ] Open Edit > Match, pick a search result, then close the modal / switch to another section / use hub back on mobile / use prev-next: confirm dialog appears; Cancel stays, Discard leaves and returns Match to the search view
- [ ] Same flows with no match selected: no dialog
- [ ] Submit a match, then leave: no dialog
- [ ] Standalone Match modal (context menu > Match) unchanged

### DIFF
--- a/src/components/modals/LibraryItemMetadataEditModal.tsx
+++ b/src/components/modals/LibraryItemMetadataEditModal.tsx
@@ -39,7 +39,7 @@ function isBookWithAudioTracks(item: BookLibraryItem | PodcastLibraryItem | null
 }
 
 /**
- * Only one section (details/chapters) is mounted at a time, so at most one of these handles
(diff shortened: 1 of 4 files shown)


mikiher:
Thanks!

Issue #267, opened 2026-09-03
Fetching a large number of episodes causes client to crash


mikiher:
While fetcing a large number of episode (>1000) from a podcast RSS feed, the browser crashed (Aw snap...) a couple of times. 

This didn't happen on one specific page, so I guess there's a memory leak somewhere around the task socket updates.

*Note: this is certainly a client bug - the server continued to work on the fetch without interruption*

mikiher:
Not sure if this is relaed, but I've seen the following at the server log.

```
(node:36608) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 drain listeners added to [Gzip]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
```

Update: this doesn't seem related. It happens when I try to add an episode from the new 1000+ episodes podcast to a playlist. This may be a different bug (related to [server bug 5524](https://github.com/advplyr/audiobookshelf/issues/5524))

PR #277, opened 2026-09-10
Discussion: Reduce item-page update overhead during batch podcast downloads

Neon1201:
﻿During large podcast downloads, each completed episode produces a full item snapshot. Applying every snapshot on the item page can create substantial update and memory overhead.

This draft coalesces item-page updates into 200ms intervals, applies the latest pending snapshot after events stop, clears pending work on item changes or unmount, and removes full-object debug logging. It does not change the socket protocol. Isolated metadata updates may appear up to approximately 200ms later under normal scheduling.

## Verification

- ESLint on the changed files passed with no findings.
- `pnpm typecheck` (`tsc --noEmit` and the Cypress `tsconfig`) passed.
- 11 Cypress component tests passed: 4 in `latestSnapshotThrottle.cy.ts`, 4 in `libraryItemUpdatedUtils.cy.ts`, and 3 in `useThrottledLatestSnapshot.cy.tsx`.
- Local production build, real Find Episodes UI, 1200-episode short-description fixture:
  - The unthrottled baseline was stopped on purpose at about 957 episodes with JS heap around 1.5GB. That was not a browser crash.
  - The 200ms candidate completed 1200 episodes in two earlier runs, with peak heap around 210-292MB, an empty download queue, and no manual refresh.
- Those numbers are local observations, not a memory cap or a general guarantee.
- The original issue's Aw Snap was not reproduced. The overhead we saw was mainly on the item page.

## Discussion

I'd appreciate feedback on whether coalescing item-page updates is an acceptable small improvement. A separate experimental approach using negotiated incremental episode events reduced socket traffic substantially, but it requires coordinated server and client changes and is not included in this PR. Would that be a useful follow-up direction?

Related to #267

### DIFF
--- a/cypress/tests/components/hooks/useThrottledLatestSnapshot.cy.tsx
+++ b/cypress/tests/components/hooks/useThrottledLatestSnapshot.cy.tsx
@@ -1,0 +1,86 @@
+import { useThrottledLatestSnapshot } from '@/hooks/useThrottledLatestSnapshot'
+import { useCallback, useState } from 'react'
+
+type Snapshot = { id: string; n: number }
+
+function ItemUpdateHarness({ itemId, intervalMs = 200 }: { itemId: string; intervalMs?: number }) {
+  const [applied, setApplied] = useState<number | null>(null)
+  const apply = useCallback((snapshot: Snapshot) => {
+    setApplied(snapshot.n)
+  }, [])
+  const schedule = useThrottledLatestSnapshot(itemId, apply, intervalMs)
+
+  return (
+    <div>
+      <span data-cy="item-id">{itemId}</span>
+      <span data-cy="applied">{applied === null ? 'none' : String(applied)}</span>
+      <button data-cy="push-1" onClick={() => schedule({ id: itemId, n: 1 })}>
+        Push 1
+      </button>
+      <button data-cy="push-2" onClick={() => schedule({ id: itemId, n: 2 })}>
+        Push 2
+      </button>
+      <button data-cy="push-stale" onClick={() => schedule({ id: 'other-item', n: 99 })}>
+        Push stale
+      </button>
+    </div>
+  )
+}
+
+function SwitchableItemHarness() {
+  const [itemId, setItemId] = useState('item-a')
+  const [applied, setApplied] = useState<number | null>(null)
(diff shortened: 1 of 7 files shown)


mikiher:
Just for clarity and so we understand what you did - do you have a root cause analysis of the issue?
If you suspect that this is due to heavy load of item updates, can you prove it?
At the least, can you prove that item updates arrive at a rate of more than one per 200 ms during a many-episode download?

Neon1201:
@mikiher Yes. I treated this as item-page update load during a many-episode download, not as a confirmed root cause of the original Aw Snap.

On the current server, each completed episode download emits a full `item_updated` snapshot (`toOldJSONExpanded()`, including the whole `episodes` array). The item page applies that snapshot on every event. That is the load this PR coalesces.

To answer the 200ms question directly: I stayed on a newly created empty podcast item page and downloaded 30 short-description episodes. Arrival was recorded **before** the 200ms throttle, in `useItemPageSocket`. A second instrument captured the same `item_updated` WebSocket frames.

Both agreed:

- 30 `item_updated` events (one per completed episode)
- 29 / 29 intervals **< 200ms (100%)**
- average 90.7ms, median 88.7ms, p95 100.8ms, min 82.2ms, max 104.0ms
- 11.02 events/sec overall, peak 12 events/sec in a 1s window
- elapsed 2.63s
- every frame included `episodes[]`; size grew from 4.8KB to 107.7KB

So during this download, item updates arrived about **11 times per second**, not once per 200ms. A 200ms latest-snapshot throttle would keep the last pending snapshot and drop the intermediate applies.

This is local (tiny audio, short descriptions). It shows arrival rate and growing full snapshots. It does not reproduce or claim to fix the original Aw Snap. This PR still only coalesces item-page applies; the socket still delivers every full snapshot.

mikiher:
Thanks for the clarification. 

Let me think about this a bit more. 

I think the item page should be able to handle updates at the rate of 12 per second, but not if each one contains a list of potentially 2000 episodes. Perhaps a server change is in order here. I'll have to think how to do this in a way that does not break the API.

Neon1201:
Yeah, I agree. It seems like a pretty complex issue and there may be several different factors involved.

PR #278, opened 2026-09-12
Move preference cookie helpers into cookies.ts

hypnotoad08:
## Summary
This PR consolidates preference-related cookie logic to reduce duplication and ensure consistent behavior between server and client writes.

## Changes
- Centralized all preference cookie names and option objects in `src/lib/cookies.ts`.
- Replaced duplicated option definitions in `lib/api.ts`, `proxy.ts`, and the theme/language routes with `PREFERENCE_COOKIE_OPTIONS`.
- Introduced a unified `COOKIE_NAMES` map covering all preference cookies, including the cover-size cookies previously defined in `lib/coverSizes.ts`.
- Updated `CardSizeContext` to use the shared `writePreferenceCookie` helper instead of its own `writeCookie`, keeping browser writes aligned with server behavior.
- Left auth cookies (`access_token`, `refresh_token`, `auth_method`) unchanged.

## Rationale
Centralizing cookie definitions reduces repetition, simplifies future updates, and ensures consistent behavior across the codebase.


advplyr:
This looks good but it has me questioning whether having a separate cover size for mobile was a good decision.

I'm going to look more into that because I think we should figure out how to do this with a single cookie. Which may require doing cover size differently than we did in Vue.

advplyr:
As far as this PR goes, the centralized cookies are working well. I'll look into the mobile cover size separately.

Thanks!

mikiher:
> This looks good but it has me questioning whether having a separate cover size for mobile was a good decision.
> 
> I'm going to look more into that because I think we should figure out how to do this with a single cookie. Which may require doing cover size differently than we did in Vue.

This can be changed to a single setting, however note that the reason that I set 100 as the default (and the maximum) for mobile is that 120 does not look good - it doesn't allow even 2 books per row on bookshelf pages on 360-px-wide screens. Anything above 100 seems to be too big. 

The mobile app allows for 2 books per row because the book size is about 100 or close.

I would actualy be fine with one single fixed size on mobile (roughly 100), and not even show the cover size widget. Larger sizes don't show well, and with smaller sizes, the text is too small to read - so the choice doesn't provide you with better options.

advplyr:
> I would actualy be fine with one single fixed size on mobile (roughly 100), and not even show the cover size widget. Larger sizes don't show well, and with smaller sizes, the text is too small to read - so the choice doesn't provide you with better options.

Agree, this is better. I pushed it: https://github.com/audiobookshelf/audiobookshelf-client-react/pull/280

Issue #258, opened 2026-08-30
Incorrect chapter label displayed at chapter boundaries


hypnotoad08:
When playback reaches or seeks to the exact start of a chapter, the player can briefly display the previous chapter’s label instead of the chapter that is starting.

Chapter timestamps are normalized to six decimal places, but the live playback position can contain additional floating-point precision. For example, a chapter may start at `12584.156112`, while the playback position is reported as `12584.156111999999.` A direct comparison treats that position as being just before the boundary and selects the previous chapter.

**Steps to reproduce**
1. Play an audiobook with chapter boundaries that align with audio file boundaries.
2. Seek to a chapter’s start or allow playback to cross the boundary naturally.
3. Observe the chapter label in the player.

**Expected behavior**

At a chapter boundary, the player should display the chapter beginning at that timestamp.

**Actual behavior**

The player may display the preceding chapter because the playback position and stored chapter boundary differ slightly due to floating-point precision.

mikiher:
Hey, I'm sorry but before I look at the PR that fixes this. I want to make sure I understand the issue - right now I'm afraid I don't.

> The player may display the preceding chapter because the playback position and stored chapter boundary differ slightly due to floating-point precision.

For how long? until the next player tick?

I tried your repro steps on a number of multi-track books and I could never reproduce this even once. You have to be more precise about the repro steps and what is it exactly that you're fixing.

hypnotoad08:
Hi, sorry, here are some screen recordings that show the issue I am seeing.

Using the main branch showing the issue: https://vimeo.com/1222490695?share=copy&fl=sv&fe=ci

While re-testing with fresh eyes this morning, I realized that this change alone does not wholly fix the issue; it initially shows correctly, but #260 shows more easily - https://vimeo.com/1222491165?share=copy&fl=sv&fe=ci

Both this fix and the fix for #260 applied - https://vimeo.com/1222491564?share=copy&fl=sv&fe=ci

mikiher:
Look, I acknowledge that #260 is an issue, but I'm also almost sure that what you're seeing here is due to the same root cause. 

It is best reproduced when you turn on "Use chapter track" and hit Next chapter a number of times. You sometimes see the full chapter track bar (signifying you're at the end of the chapter) flash for a second or less, and then it becomes the new (empty) chapter bar. 

You said: 

> While re-testing with fresh eyes this morning, I realized that this change alone does not wholly fix the issue

[...]

PR #279, opened 2026-09-13
Fix seeking and chapter navigation across track boundaries

### LINKED ISSUE #260: Seeking across MP3 file boundaries can go to the wrong position
Seeking can behave incorrectly in books made up of multiple MP3 files. This affects jump buttons, chapter selection, and any seek that lands exactly at the point where one file ends and the next begins. At exact boundaries, the player may choose the wrong file, and during a file change, it can briefly report an incorrect book position while the new file loads.

## Steps to reproduce

1. Play a book made up of multiple MP3 files.
2. Move close to the end of one file.
3. Use the forward or backward jump button to cross into the next or previous file.
4. Try the same while playing and while paused.
5. Seek to a chapter that starts exactly at a file boundary.
6. Repeat navigation while the destination file is still loading.

## Expected behavior

- The player selects the correct MP3 file.
- It seeks to the requested position.
- It reports a consistent book position while the new file loads.
- It stays in the correct playing or paused state.

## Actual behavior

- At exact boundaries, the player may seek within the wrong file.
- During a file change, it can briefly display an incorrect time or chapter progress.
- Repeated navigation while the new file is loading can interrupt playback.

hypnotoad08:
Closes #260

## Problem

Books split across multiple audio files could behave incorrectly when seeking or navigating chapters.

- Seeking to an exact track boundary could resolve to the wrong track.
- Seeking to the exact end of a book from any track before the final track was silently ignored because no track's range contained that timestamp.
- While a cross-track seek was loading, `usePlayerHandler` stopped updating playback time and chapter state. During that window, `nextChapter` and `previousChapter` could become stale, causing chapter navigation to target the wrong chapter.
- During cross-track loads, the player could temporarily report an incorrect playback position or chapter progress.
- Seeks issued while a track was loading could be ignored.
- Repeated navigation during a track change could interrupt playback or lose the intended play/pause state.

### Chapter navigation

Added `src/lib/chapters/chapterPlayback.ts` with utilities for resolving chapter navigation from a playback position:

- `findChapterNavigationAtTime`
- `resolvePreviousTarget`
- `resolveNextTarget`

Previous and next chapter actions now resolve their target directly from the current playback position instead of relying on `usePlayerHandler`'s cached `previousChapter` and `nextChapter` state.

This avoids stale chapter navigation during cross-track loads. The existing "restart current chapter if more than 3 seconds in, otherwise go to the previous chapter" behavior is unchanged.

Updated:

- `usePlayerChapterQueueNavigation.ts`
- `usePlayerHandler.ts`

### Local audio playback

Updated `src/lib/player/LocalAudioPlayer.ts`:

- Track selection now consistently uses half-open ranges, ensuring exact boundary timestamps resolve to the following track.
- Seeking to the exact end of a book now resolves to the final track instead of being ignored.
- Added `isTrackLoading` to track in-flight track changes.
- `getCurrentTime()` now reports the requested playback position while a track is loading instead of transient audio-element state.
- Play/pause intent is preserved across cross-track loads.
- New seek requests retarget an in-flight load instead of being dropped.
- Playback now resumes only after the seek associated with a track load completes, preventing premature resumes while a seek is still pending.

## Testing

Added or expanded Cypress coverage for:

- Exact track boundaries
- Exact chapter boundaries
- Seeking to the end of a book
- Seeking into gaps between tracks
- Retargeting seeks during an in-flight load
- Preserving play/pause intent across track changes
- Previous/next chapter behavior near chapter boundaries


mikiher:
OK, I made a few changes and fixes:
- Removed the chapterPlayback helper and inlined its code (we try not to introduce new files if they're not really required).
- Deferred currentTime writes until the audio element can accept them (mostly defensive)
- Made a number of changes so that the play button and the chapter track bar do not flicker when chapters are switched
  - The last change is actually similar to your original fix to #258. I applied it after debugging and finding that sometimes next/prev chapter seeks fell less than 1ms before chapter start, and needed to be snapped. So it seems that PR #259 actually did fix an issue. Appologies for doubting that fix, but I had to see some proof of the problem, which eventually happened during debugging.

This now seems to be working well on Desktop, Chrome Android, and Safari iOS.

Thanks!

hypnotoad08:
> OK, I made a few changes and fixes:
> 
> * Removed the chapterPlayback helper and inlined its code (we try not to introduce new files if they're not really required).
> * Deferred currentTime writes until the audio element can accept them (mostly defensive)
> * Made a number of changes so that the play button and the chapter track bar do not flicker when chapters are switched
>   
>   * The last change is actually similar to your original fix to [Incorrect chapter label displayed at chapter boundaries #258](https://github.com/audiobookshelf/audiobookshelf-client-react/issues/258). I applied it after debugging and finding that sometimes next/prev chapter seeks fell less than 1ms before chapter start, and needed to be snapped. So it seems that PR [Fix chapter labels and navigation at chapter boundaries #259](https://github.com/audiobookshelf/audiobookshelf-client-react/pull/259) actually did fix an issue. Appologies for doubting that fix, but I had to see some proof of the problem, which eventually happened during debugging.
> 
> This now seems to be working well on Desktop, Chrome Android, and Safari iOS.
> 
> Thanks!

All good, glad you where able to reproduce it!

Issue #222, opened 2026-08-24
[BUG] - Podcasts missing schedule


SAS-1:
Hi

When going to Library under a Podcasts library and editing the record there is no way to see if the RSS feed is enabled or what the schedule is set to.

<img width="414" height="413" alt="Image" src="https://github.com/user-attachments/assets/85ef5f7a-05e8-4efc-bf8f-77ee4cfa4c53" />

only option seems to be going into each one individually and selecting schedule from the hamburger menu, so no way to scan through all the podcasts quickly to check the schedule or if they are enabled.

mikiher:
@advplyr did you plan on adding Schedule to Podcast sections?

advplyr:
Yeah I think it can fit in there. It's metadata on the podcast

mikiher:
While it _can_ fit there, I don't think it's the best fit. 

There are 3 modals which have to do with **fetching episodes** from the podcast's feed:
- Schedule
- Find episodes
- Check for new episodes

If you put one of them as a section in the Edit Modal, the others should be there as well (because why not?)

Except I don't think they should be there.

If you want to show them together somehere (which makes sense), I think it should be under a separate SectionedModal triggered from a separate button, that only appears when there's an RSS feed (since they all depend on the existance of the feed). That SectionedModal can have navigation like the Edit modal, but it would be dedicated to getting episodes, not editing the podcast.

I know some users will complain that they don't have all the podcast actions at the tip of their fingers like in Vue, where everything is crammed under the same tabbed modal - but I think it's not the best UX, since it groups actions that are not semantically related (it is the same as separating metadata manipulation from file tools in books), and I believe it would overload the modal (just like it is overloaded in Vue).

My 2 cents.

advplyr:
It's not exactly like separating metadata manipulation and file tools in books because the schedule is metadata that we store on the podcast record.
Having a separate sectioned modal for managing RSS feed downloads does make more sense though. Especially since we consider downloading from the RSS feed a convenience feature rather than a primary function.

Caleb68864:
I’d like to work on this issue.

I reproduced the current behavior: podcast cards expose Schedule and Check for New Episodes separately, while Find Episodes is only reached through the podcast item page.

[...]

PR #285, opened 2026-09-18
Group podcast RSS actions in a sectioned modal

Caleb68864:
**What this changes**

Schedule, Find Episodes and Check for New Episodes are now reached through a single "Podcast RSS Actions" entry that opens a `SectionedModal`, instead of separate menu entries. The entry only appears for podcasts that actually have a source RSS feed.

Following the discussion on #222, this is a dedicated modal for fetching episodes rather than a section inside the edit modal. Reusing `SectionedModal` means the desktop rail and the mobile hub, drill-in and Back behaviour all come from the existing shared component rather than being rebuilt.

Media cards and the podcast item page render the same feature, so the action behaves identically wherever it is opened.

The podcast edit view also now shows read-only RSS status: whether a source feed is configured, whether automatic fetching is enabled, and the stored schedule. That covers the part of the issue about not being able to tell what a podcast is set to without opening the schedule dialog for each one.

**Review feedback addressed**

- *"Clicking Podcast RSS Actions on the podcast item page does nothing"* — the item page set the open state but never rendered the feature, so it was a silent no-op. It now renders the same `PodcastRssActionsFeature` the cards use, so both surfaces have one owner.
- *"Clicking Find Episodes does nothing"* — it now loads the item with `include=downloads`, fetches the feed, and opens the existing `EpisodeFeedModal` with the saved episodes and current download queue, so episodes that already exist are not offered again. It also uses the feed URL from that response rather than a card that may have gone stale.
- Desktop and mobile behaviour now come from `SectionedModal`, as suggested.

**One judgement call worth your opinion**

The stored schedule is shown in the edit view even when automatic fetching is disabled, under a "Schedule (inactive):" heading so it cannot be mistaken for a running schedule. A podcast that is switched off still keeps its schedule, and hiding it answers only half of "is it enabled, or what is it set to". Happy to hide it instead if you would rather.

**Notes**

- No new API routes. The Schedule and Check for New Episodes bodies are now exported as form components and reused directly by the grouped manager, so the two entry points cannot drift apart.
- Because the card and the item page both use the grouped manager, the standalone `PodcastDownloadScheduleModal` and `PodcastCheckNewEpisodesModal` dialogs are no longer rendered anywhere. I kept the thin wrappers rather than delete them — say the word and I will remove them instead of leaving unused components behind.
- The schedule editor previously fell back to the daily default whenever automatic downloads were off, so re-enabling a podcast silently overwrote its saved cron expression. It now loads whatever is stored.
- A late feed response cannot open the episode viewer after the manager has been closed or the section has changed.
- An earlier revision of this PR also added the RSS status to library cards. That is not included here: it needs its own design discussion and it was not what the review was about.
- Bulk podcast management is deliberately out of scope.

**Testing**

- Component suite: 773 passing, 0 failing.
- TypeScript (app and Cypress), ESLint, Prettier and the hardcoded-string check all pass.
- Manually against a local server with a library of fourteen podcasts from real feeds: the manager opens from the library card, home card and item page, on desktop and at 390px; Find Episodes opens the real feed viewer from all three; Escape closes the feed viewer before the manager; a podcast with no source feed hides the action entirely; a rejected feed shows an error and leaves the manager usable; a non-admin with update permission does not see the action; and closing the manager or switching section while a feed request is in flight does not pop the viewer open afterwards.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012dr1jKjcoTHMJndEDYFAUR


mikiher:
There are a few issues with this:
1. Clicking on "Podcast RSS Actions" on the podcast item page does nothing
2. Clicking on "Find Episodes" on the selection modal does nothing

However I think the general experience should be similar to edit modal:
1. There's no value in displaying another selection modal on desktop. On mobile, it does make sense. If you use SectionedModal machinery, you'll get those behaviors for free.
2. Make sure it works both on cards (homepage/library) and on the item page

mikiher:
- The point of SectionedModal is to embed existing modals in it. The Find Episodes section should contain the modal, not a button that opens it. Use a loading indicator to show loading state.
- I don't like the RSS status in the podcast details view. It makes the details overflow and is not actionable.

Caleb68864:
Both addressed and pushed — the head is now `9041007`.

**Find Episodes embeds the browser.** I split `EpisodeFeedModal` the same way
`PodcastDownloadScheduleModal` and `PodcastCheckNewEpisodesModal` were already
split earlier in this PR: a new exported `EpisodeFeedList` holds the search,
sort, selection and download behaviour with no dialog chrome, and the
default-exported `EpisodeFeedModal` is now a thin wrapper around it. The other
caller, `EpisodeTable`, sees an unchanged prop contract.

`FindEpisodesPanel` no longer renders a heading and a launch button. It
requests the feed on mount, shows `LoadingIndicator variant="inline"` while
that runs, and renders `EpisodeFeedList` in place — so it is one dialog now,
not two. Mount means active here: `Modal` renders nothing while closed and
`SectionedModalBody` renders no children on the mobile hub, so the request
starts exactly when the user reaches the section and leaving it unmounts the
panel. Empty and failed feeds are reported inside the panel now, with a Retry,
rather than as a toast over an empty section.

**The RSS status is out of the details view.** `LibraryItemEditModal.tsx` is
back to its upstream content — `git diff origin/master` on that file is empty.
`PodcastRssStatusDetails.tsx`, `podcastRssStatus.ts`, their two Cypress specs
and the six locale keys only they used are deleted. I left the removal as its
own commit rather than force-pushing over the commit that added it, so the
response to your comment stays visible in the history — happy to collapse the
two if you would rather the branch read clean.

On coverage, so it is not overstated: the component suite is 763 passing (down
from 773 — the two deleted status specs carried ten tests). The embedded panel
itself has **no** component test. The Cypress component build cannot mount
anything that imports a server action — webpack fails with *"You're importing a
module that depends on `next/headers`"* from `src/lib/api.ts` — and no existing
spec does it either, which is why the navigator spec injects a fake
`renderSection`. I verified the panel in a browser instead: loading indicator,
then the embedded browser in a single dialog, the rail still switching to
Schedule while it is embedded, the mobile drill-in keeping Back without
overflowing 390px, and a rejected feed showing the inline error with a working
Retry. If you want that gap closed properly it means aliasing the action module
in the Cypress webpack config, which I am glad to do as its own PR rather than
growing this one.

One thing I did not decide on my own: #222 asked to see "if the RSS feed is
enabled or what the schedule is set to", and removing the status block leaves
that answerable only from inside the manager. I have not tried to slip a
replacement indicator in somewhere else. If a passive one is wanted, I will
propose it separately.

One ask: the CI run is sitting at `action_required` rather than failed — fork
PRs need a maintainer to approve the workflow run, and this push reset the
approval the earlier green `quality` check had, so nothing has run on
`9041007` yet. Could you hit "Approve and run workflows" when you next look?
Happy to fix whatever it turns up.

mikiher:
I'd like @advplyr to approve UI/UX before I continue the code review.

One thing I don't like is that the section side buttons have too much text ("Find Episodes" and "Check for new Episodes"). I think it needs to be reduced to one word and should not wrap if possible (certainly not wrap twice). Not sure exactly how to resolve this, maybe replace with "Find/Lookup" and "Check New".

Caleb68864:
Understood on waiting for @advplyr before the rest of the review — no rush from my side.

I pushed the label change in `dfdb2bd` so there is something concrete to look at.

The rail is one icon wide (`w-20`), so any label longer than a word wraps there. Rather than shortening the wording everywhere, `Section` now takes an optional `shortLabel` that **only the desktop rail** uses; the mobile hub rows and the drill-in header are full width, so they keep the descriptive text. The rail button also carries the full label as its `title`, so hovering still spells the action out.

Desktop rail: **Schedule** / **Lookup** / **Check** — three single-line labels, no wrapping.
Mobile hub: Schedule / Find Episodes / Check for new episodes, unchanged.

On the wording: "Lookup" is the existing `ButtonLookup` key, so it is already translated in every locale and it is the word the Vue client uses for the same "go read the feed" action. "Check" is the one new en-us key (`ButtonCheck`). If you would rather not add a key at all, `ButtonRefresh` ("Refresh") exists and matches the section's refresh icon — I avoided it because it reads like "reload the view" rather than "fetch new episodes". Your "Check New" also works but is two words, which wraps once at 80px. Happy to take whichever you and @advplyr prefer; it is a one-line change.

I also gave the rail label `w-full px-1 text-center leading-tight` so a long translation of any section label in any sectioned modal degrades rather than overflowing the rail.

The spec now asserts the short label and the tooltip. `PodcastRssActionsModal.cy.tsx` passes, as does the rest of the modal suite, along with typecheck (app and Cypress), ESLint, Prettier and the hardcoded-string check.

Still standing: the CI run needs a maintainer to hit "Approve and run workflows" — nothing has run on this branch since the approval reset.

advplyr:
Thanks for contributing. I'm wondering if there is a human in the loop here at all. Was the output of these changes visually reviewed by a human or is this entirely automated?

PR #286, opened 2026-09-20
Update player progress bar with chapter segments

Neon1201:
## Summary

This updates the audiobook player progress bar to represent chapters as interactive segments.

- show chapter boundaries as circular markers inside the progress track
- highlight individual chapter ranges on hover and focus
- display the chapter number, title, and start time
- click a boundary marker to jump to the chapter start
- preserve normal click-to-seek behavior within the progress bar

## Details

Chapter segments are calculated from each chapter start to the next, with the final segment ending at the media duration. The first chapter remains selectable without displaying a redundant boundary marker at 00:00.

Invalid or out-of-range chapter timestamps are ignored. Audiobooks without chapters retain the existing progress bar design and behavior.

## Accessibility

Chapter boundary markers support keyboard focus and Enter/Space activation. Chapter information is available through accessible labels and tooltips.

## Testing

- Cypress utility tests for chapter validation, ordering, labels, and segment ranges
- Cypress component tests for rendering and interactions
- exact progress-bar seeking
- chapter-boundary seeking
- keyboard interaction
- desktop and mobile layouts
- no-chapter fallback
- ESLint and TypeScript checks
- production build

## Screenshot
d27cae84-8918-4e4b-ab6c-bd53b6aa4c75


mikiher:
@Neon1201 thanks for putting out this PR, but at this point I'd like to shelf it until we release audiobookshelf-client-react.

Since I think we're about 1-2 months from release, we'd like to focus only on bug fixes and must-have features that @advplyr or me approve. We both have limited review cycles, and while we really appreciate contributions, we have to focus only on PRs that will advance us towards this milestone.

I'll close this for now, but please open a feature request and keep the code around for when we approve it. 

Thanks for understanding!

